PR #14, opened 2020-01-16
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action

One limitation currently exists. This action doesn't support nightly releases.
If you believe that you will ever need to depend upon nightly versions, we'll have an update shortly that allows for this (but this will require changes to `services.gradle.org`).


breskeby:
Nice!

PR #18, opened 2020-01-22
improved spock parameter detection

jonnybbb:


jkschneider:
Thanks @jonnybbb!

breskeby:
@jonnybbb did you test this already with the GE build?

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
```diff
@@ -36,6 +37,11 @@
 public class SpockParameterClassVisitor extends ClassVisitor {
 
     private static final Set<Character> REGEX_CHARS = setOf(new char[]{'<', '(', '[', '{', '\\', '^', '-', '=', '$', '!', '|', ']', '}', ')', '?', '*', '+', '.', '>'});
+    //A valid Java identifier https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8 including methods
```
JLLeitschuh:
```suggestion
// A valid Java identifier https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8 including methods
```

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
```diff
@@ -58,10 +64,11 @@ public MethodVisitor visitMethod(int access, String name, String descriptor, Str
     public void visitEnd() {
         spockMethodVisitor.getTestMethodPatterns().stream()
             .filter(methodPattern -> {
-                String methodPatternRegex = escapeRegEx(methodPattern).replaceAll("#\\w+", "\\\\w+");
+                //detects a valid spock parameter and replace it with a wildcards http://spockframework.org/spock/docs/1.3/data_driven_testing.html#_more_on_unrolled_method_names
```
JLLeitschuh:
```suggestion
// detects a valid spock parameter and replace it with a wildcards http://spockframework.org/spock/docs/1.3/data_driven_testing.html#_more_on_unrolled_method_names
```

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
```diff
@@ -36,6 +37,11 @@
 public class SpockParameterClassVisitor extends ClassVisitor {
 
     private static final Set<Character> REGEX_CHARS = setOf(new char[]{'<', '(', '[', '{', '\\', '^', '-', '=', '$', '!', '|', ']', '}', ')', '?', '*', '+', '.', '>'});
+    //A valid Java identifier https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8 including methods
+    public static final String SPOCK_PARAM_PATTERN = "#[\\p{L}\\d$_.()&&[^#\\s]]+";
+    public static final String PARAM_PLACEHOLDER = "#param";
+    public static final String WILDCARD_SUFFIX = ".*";
+    public static final String WILDCARD = WILDCARD_SUFFIX;
```
guylabs: Why is that constant required, as it's the same value as `WILDCARD_SUFFIX`?

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
```diff
@@ -58,10 +64,11 @@ public MethodVisitor visitMethod(int access, String name, String descriptor, Str
     public void visitEnd() {
         spockMethodVisitor.getTestMethodPatterns().stream()
             .filter(methodPattern -> {
-                String methodPatternRegex = escapeRegEx(methodPattern).replaceAll("#\\w+", "\\\\w+");
+                //detects a valid spock parameter and replace it with a wildcards http://spockframework.org/spock/docs/1.3/data_driven_testing.html#_more_on_unrolled_method_names
+                String methodPatternRegex = escapeRegEx(methodPattern.replaceAll(SPOCK_PARAM_PATTERN, PARAM_PLACEHOLDER)).replaceAll(PARAM_PLACEHOLDER, WILDCARD) + WILDCARD_SUFFIX;
```
guylabs: I'd extract the `methodPattern.replaceAll(SPOCK_PARAM_PATTERN, PARAM_PLACEHOLDER)` part to a `normalizeMethodName()` method to make it more readable.

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
jonnybbb: just for better readability

guylabs: I checked twice when reading the regex replacement ;). But lets keep it as it is.

PR #22, opened 2020-01-28
Rerun a parameterized test of an abstract class within a test hierarchy

### LINKED ISSUE #21: cannot rerun failed tests of super class with parameterized test method
Within a test hierarchy a parameterized test method of a super class cannot be rerun in case of failure. 

```
 @spock.lang.Unroll
  abstract class AbstractTest extends spock.lang.Specification {
          def "unrolled [#param] parent"() {
               expect:
               ${flakyAssert()}

               where:
               param << ['foo']
                }
            }
            
class InheritedTest extends AbstractTest {

           def "inherited"() {
               expect:
               true
            }
         }

```

results in 
```
* What went wrong:
Execution failed for task ':test'.
> org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues
     acme.InheritedTest#unrolled [foo] parent
```

jonnybbb:
Fixes #21

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/RetryTestFrameworkGenerator.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/RetryTestFrameworkGenerator.java
@@ -151,7 +151,7 @@ final class RetryTestFrameworkGenerator {
                     .map(testClass -> {
                         try (FileInputStream testClassIs = new FileInputStream(testClass)) {
                             ClassReader classReader = new ClassReader(testClassIs);
-                            SpockParameterClassVisitor visitor = new SpockParameterClassVisitor(failedTest.getName());
+                            SpockParameterClassVisitor visitor = new SpockParameterClassVisitor(failedTest.getName(),spec);
                             classReader.accept(visitor, 0);
                             return new TestName(failedTest.getClassName(), visitor.getTestMethodName());
                         } catch (Throwable t) {
--- a/plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
@@ -15,16 +15,26 @@
  */
 package org.gradle.testretry.internal.visitors;
 
+import org.gradle.api.file.FileCollection;
+import org.gradle.api.internal.tasks.testing.JvmTestExecutionSpec;
 import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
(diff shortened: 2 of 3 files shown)


plugin/src/main/java/org/gradle/testretry/internal/RetryTestFrameworkGenerator.java
```diff
@@ -151,7 +151,7 @@ private static boolean isSpockStepwiseTest(JvmTestExecutionSpec spec, TestName f
                     .map(testClass -> {
                         try (FileInputStream testClassIs = new FileInputStream(testClass)) {
                             ClassReader classReader = new ClassReader(testClassIs);
-                            SpockParameterClassVisitor visitor = new SpockParameterClassVisitor(failedTest.getName());
+                            SpockParameterClassVisitor visitor = new SpockParameterClassVisitor(failedTest.getName(),spec);
```
jonnybbb: TestName extraction should go into yhe SpockParameterClassVisitor completly

plugin/src/main/java/org/gradle/testretry/internal/visitors/SpockParameterClassVisitor.java
```diff
@@ -80,6 +95,25 @@ private static String normalizeMethodName(String methodPattern) {
         return methodPattern.replaceAll(SPOCK_PARAM_PATTERN, PARAM_PLACEHOLDER);
     }
 
+    @Override
+    public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
+        super.visit(version, access, name, signature, superName, interfaces);
+        if (superName != null && !superName.equals("java/lang/Object")) {
+            final FileCollection collection = spec.getTestClassesDirs().filter(d -> Files.exists(parentClass(d, superName)));
```
jonnybbb: can we even make better use of the gradle JvmTestExecutionSpec (eg `JvmTestExecutionSpec.getPreviousFailedTestClasses()`). The `org.gradle.api.internal.tasks.testing.junit.JUnitDetector` knows already about the superClasses

ldaley:
Solution needs to be more general. I'm going to close this one out. We can still use the code as guidance in the more general solution.

jkschneider:
> Solution needs to be more general.

In fact this isn't necessary. Spock is the only framework for which we need to resort to ASM to look at superclass hierarchy. Added tests for all other frameworks in 92ae3adfe5da93a63146dc643055e50931dec5a3.

@jonnybbb had identified in one of the commits the desire to generalize how to locate the bytecodes for a Spock test's base class (from assuming it is in the same test source set to retrieving it from other places in the classpath). This was achieved in f5b5978180cb351e11d67df799dfa15743c37e97.

PR #26, opened 2020-02-04
Handle Spock names that also match a parameterized test name

### LINKED ISSUE #25: Failure when retrying Spock test when `failOnPassedAfterRetry = true`
I got `org.gradle.test-retry was unable to retry the following test methods, which is unexpected` when retrying `net.rubygrapefruit.platform.file.WindowsFileEventFunctionsTest#can detect file removed`. See https://github.com/gradle/native-platform/blob/master/build.gradle#L19-L25 and https://e.grdev.net/s/72ekrn4it7gpc/failure?openFailures=WzBd&openStackTraces=WzEsMF0#top=0.

jkschneider:
Fixes #25. Also, we're going to be more pessimistic about matching now. If a failed name matches more than one name on a Spock class for whatever reason, we'll just retry all of them. No more picking winners, except when the failed name exact matches a test name.


wolfs:
Did you confirm it fixes the problem for the `native-platform` project?

jkschneider:
> Did you confirm it fixes the problem for the `native-platform` project?

Yes, I did.

PR #41, opened 2020-04-06
Allow to deactivate replacing the TestExecuter

marcphilipp:
For interoperability with the test-distribution plugin, the test-retry
plugin now calls the `shouldTestRetryPluginBeDeactivated()` method on
the "distribution" extension. If it returns true, it does not replace
the `TestExecuter` since that will be done by the test-distribution
plugin.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/TestTaskConfigurer.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/TestTaskConfigurer.java
@@ -21,7 +21,6 @@ import org.gradle.api.internal.initialization.loadercache.ClassLoaderCache;
 import org.gradle.api.internal.tasks.testing.JvmTestExecutionSpec;
 import org.gradle.api.internal.tasks.testing.TestExecuter;
 import org.gradle.api.model.ObjectFactory;
-import org.gradle.api.plugins.ExtraPropertiesExtension;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/org/gradle/testretry/internal/TestTaskConfigurer.java
```diff
@@ -90,12 +91,34 @@ private static boolean supportsPropertyConventions(VersionNumber gradleVersion)
...
+        public ConditionalTaskAction(Action<Test> delegate) {
+            this.delegate = delegate;
+        }
 
         @Override
         public void execute(@NotNull Task task) {
-            Test testTask = (Test) task;
-            TestExecuter<JvmTestExecutionSpec> testExecuter = getTestExecuter(testTask);
+            ExtraPropertiesExtension extra = task.getExtensions().getExtraProperties();
+            boolean deactivated = extra.has(DEACTIVATION_PROPERTY_NAME) && Boolean.parseBoolean(String.valueOf(extra.get(DEACTIVATION_PROPERTY_NAME)));
```
ldaley, line 107: What's this about? I don't see any mention of it in the spec, and don't understand what it's for.

marcphilipp: Resolved in other channels.

PR #56, opened 2020-07-29
Update to use Gradle Enterprise plugin v3.4

bigdaz:
Also move the `buildScan` configuration from the root build scriptinto `settings.gradle.kts`.


jonnybbb:
@bigdaz I applied your changes on #60

Issue #52, opened 2020-06-26
Test that are skipped after failing are considered to have passed


ldaley:
This is unlikely to match the user's intention.

Such a scenario should not be considered success. Whether tests that are skipped after failure should be retried until retries are exhausted is a separate question. It might be better to stop as soon as this situation is encountered.

marcphilipp:
> It might be better to stop as soon as this situation is encountered.

I tend to agree.

jonnybbb:
So the build should fail for a case where a test would `FAIL` in the first run. `SKIPPED` in the first retry and finally would have `PASSED` in the second retry run, since the retry should better stop in such a situation?

marcphilipp:
I think so but we should retry other failed tests and fail in the end.

jonnybbb:
Keep retrying `FAIL` and then `SKIPPED` tests - seems to me a simpler approach. @ldaley / @marcphilipp I am not sure if we should stop retrying those tests in such a situation. Can you give me some more context why retrying should be stopped?

ldaley:
@jonnybbb “simpler” for who? If you mean in terms of implementation, that's the wrong perspective. 

I can't imagine a situation where we would get FAILED, SKIPPED, PASSED. So, why continue? That said, I don't feel strongly. I can buy the argument that treating SKIPPED as FAILED in a retry context presents “simpler” behaviour to users. Let's just go with that.

jonnybbb:
My intention was not to propose a "simpler" solution at the expense of the "right" solution. Depending on how `@IgnoreIf` or `assumeThat()` that result in SKIPPED test is used, there might be rare cases out there resulting in such a FAILED, SKIPPED, PASSED scenario. 
I agree with your proposal. The PR https://github.com/gradle/test-retry-gradle-plugin/pull/59 is ready for review.

PR #59, opened 2020-08-02
Build should fail on a FAILED/SKIPPED test scenario 

jonnybbb:
Testcases to reproduce the issue of #52 and the actual fix for it.


plugin/src/main/java/org/gradle/testretry/internal/RetryTestResultProcessor.java
```diff
@@ -68,11 +68,20 @@ public void started(TestDescriptorInternal descriptor, TestStartEvent testStartE
 
     @Override
     public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
-        activeDescriptorsById.remove(testId);
+
+        final TestDescriptorInternal descriptor = activeDescriptorsById.get(testId);
```
marcphilipp, line 72: nit: we don't seem to be consistent in our use of `final` in this class. We should either always use it when possible, or never.

marcphilipp:
```suggestion
        final TestDescriptorInternal descriptor = activeDescriptorsById.remove(testId);
```

plugin/src/main/java/org/gradle/testretry/internal/RetryTestResultProcessor.java
```diff
@@ -68,11 +68,20 @@ public void started(TestDescriptorInternal descriptor, TestStartEvent testStartE
...
+            final boolean hasFailedPreviously = nonExecutedFailedTests.remove(test);
+            if (hasFailedPreviously && testCompleteEvent.getResultType() == SKIPPED) {
+                failedTests.add(test);
+            }
+        }
 
         if (!testId.equals(rootTestDescriptorId) || lastRun()) {
             delegate.completed(testId, testCompleteEvent);
         }
+        activeDescriptorsById.remove(testId);
```
marcphilipp, line 84:
```suggestion
```

plugin/src/main/java/org/gradle/testretry/internal/RetryTestResultProcessor.java
```diff
@@ -84,7 +93,9 @@ public void output(Object testId, TestOutputEvent testOutputEvent) {
     public void failure(Object testId, Throwable throwable) {
         TestDescriptorInternal descriptor = activeDescriptorsById.get(testId);
         if (descriptor != null && descriptor.getClassName() != null) {
-            failedTests.add(new TestName(descriptor.getClassName(), descriptor.getName()));
+            final TestName test = new TestName(descriptor.getClassName(), descriptor.getName());
+            nonExecutedFailedTests.remove(test);
```
marcphilipp, line 97: ❓ Why are we removing it here and in `completed`?

plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
```diff
@@ -647,6 +648,130 @@ class SpockFuncTest extends AbstractPluginFuncTest {
...
+
+            @spock.lang.Stepwise
+            class StepwiseTests extends spock.lang.Specification {
+                def "parentTest"() {
+                    expect:
+                    true
+                }
+
+                @spock.lang.IgnoreIf({Files.exists(Paths.get("build/marker.file")) })
+                def "childTest"() {
```
marcphilipp, line 672: ❓ Why "child" and "parent"?

plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
```diff
@@ -647,6 +648,130 @@ class SpockFuncTest extends AbstractPluginFuncTest {
...
+        result.output.count('a flaky Test PASSED') == 0
+        result.output.contains('4 tests completed, 1 failed, 3 skipped')
+        !result.output.contains('Please file a bug report at')
+
+        where:
+        gradleVersion << GRADLE_VERSIONS_UNDER_TEST
+    }
+
+    @Unroll
+    def "build is successful if a test is ignored but never passed (gradle version #gradleVersion)"() {
```
marcphilipp, line 734:
```suggestion
    def "build is successful if a test is ignored but never failed (gradle version #gradleVersion)"() {
```

plugin/src/main/java/org/gradle/testretry/internal/RetryTestResultProcessor.java
jonnybbb: previously it was executed beforehand in the `started` method which is called on every test execution. 
https://github.com/gradle/test-retry-gradle-plugin/pull/59/files#diff-c3de119e7a9886810b316ca4173f467eL55
But you are right its needless in the FAILED case,

plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
jonnybbb: The parent is indeed not required here anymore.

plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
```diff
@@ -661,15 +661,10 @@ class SpockFuncTest extends AbstractPluginFuncTest {
             import java.nio.file.Paths
             import java.nio.file.Files
 
-            @spock.lang.Stepwise
             class StepwiseTests extends spock.lang.Specification {
```
marcphilipp, line 664: This is no longer a "stepwise" test.

PR #60, opened 2020-08-02
Replace deprecated internal Test.getClassLoaderCache() 

### LINKED ISSUE #53: Internal Test.getClassLoaderCache() deprecated in 6.6, will be removed in 7.0
This plugin relies on `Test.getClassLoaderCache()` which has been deprecated for Gradle 6.6 and  will be removed in 7.0.
This should be fixed in order to not emit a deprecation warning on Gradle >= 6.6 and not break on Gradle >= 7.0.

jonnybbb:
Fixes #53

Checklist:
- [x] add PR for BT https://github.com/gradle/gradle/pull/13993 to annotate used internal classes with @UsedByScanPlugin("test-retry")
- [x] add daily build to CI Pipeline


samples/sample-junit5/build.gradle.kts
```diff
@@ -1,7 +1,8 @@
 plugins {
     java
     jacoco
-    id("org.gradle.test-retry") version "1.0.0"
+    //TODO this needs to replaced as soon as 1.1.7 is out together with junit5.sample.out#L9 and remove `--warning-mode all` from junit5.sample.confL2
+    id("org.gradle.test-retry") version "1.1.6"
```
marcphilipp, line 5: ❓ Why do we need to declare a version here? Shouldn't we inject to the `GradleRunner` via the classpath?

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
```diff
@@ -59,7 +61,7 @@ public TestFramework createRetrying(JvmTestExecutionSpec spec, Test testTask, Se
                 }
             });
 
-        return new TestNGTestFramework(testTask, retriedTestFilter, instantiator, classLoaderCache);
+        return new TestNGTestFramework(testTask, ((ProjectInternal)testTask.getProject()).getServices().get(ObjectFactory.class).fileCollection(), retriedTestFilter, ((ProjectInternal)testTask.getProject()).getServices().get(ObjectFactory.class));
```
marcphilipp, line 64: ❓ Why was this change necessary? `ObjectFactory` could be extracted as a local variable to increase readability.

samples/sample-junit5/build.gradle.kts
```diff
@@ -1,7 +1,8 @@
 plugins {
     java
     jacoco
-    id("org.gradle.test-retry") version "1.0.0"
+    //TODO this needs to replaced as soon as 1.1.7 is out together with junit5.sample.out#L9 and remove `--warning-mode all` from junit5.sample.confL2
```
marcphilipp, line 4: Rather than a TODO comment, should we rather create an issue so we won't forget?

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
jonnybbb: This change was necessary to be compatible with Gradle 6.6. But it actually breaks backwards compatibility with previous Gradle versions. I think we should aim for backwards compatibility here and extend the implementation in that regard.

samples/sample-junit5/build.gradle.kts
jonnybbb: I propose to change the sample build so that we can test with multiple versions and inject the version via the `GradleRunner` as you suggested and then we do not need the TODO anymore.

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
marcphilipp: I agree. We should add `@UsedByScanPlugin("test-retry")` annotations to all internal APIs that we rely on so this doesn't go unnoticed next time.

samples/sample-junit5/build.gradle.kts
jonnybbb: I am not sure how to apply the `GradleRunner` in that context. Its a sample project executed via [Examplar](https://github.com/gradle/exemplar) - Nevertheless the latest test-retry-plugin is now always applied.

samples/sample-junit5/build.gradle.kts
jonnybbb: I mixed up `GradleSamplesRunner` with `GradleRunner` in your comment. I could get rid of the TODO by applying some templating to fetch the latest plugin version.

samples/sample-junit5/junit5.sample.out
```diff
@@ -7,17 +7,18 @@
...
 
-SampleTests > failing() FAILED
-    org.opentest4j.AssertionFailedError at SampleTests.java:21
+org.gradle.testretry.sample.SampleTests > failing() FAILED
+    org.gradle.internal.serialize.PlaceholderAssertionError at SampleTests.java:21
 
-SampleTests > failing() FAILED
-    org.opentest4j.AssertionFailedError at SampleTests.java:21
+org.gradle.testretry.sample.SampleTests > failing() FAILED
+    org.gradle.internal.serialize.PlaceholderAssertionError at SampleTests.java:21
```
jonnybbb, line 21: @marcphilipp is the internal class `org.gradle.internal.serialize.PlaceholderAssertionError` showing up here ok?

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
```diff
@@ -59,7 +64,19 @@ public TestFramework createRetrying(JvmTestExecutionSpec spec, Test testTask, Se
                 }
             });
 
-        return new TestNGTestFramework(testTask, retriedTestFilter, instantiator, classLoaderCache);
+        VersionNumber gradleVersion = VersionNumber.parse(testTask.getProject().getGradle().getGradleVersion());
+        if (gradleVersion.getMajor() >= 6 && gradleVersion.getMinor()>= 6) {
```
marcphilipp, line 68: Can be simplified using `org.gradle.util.GradleVersion`:
```suggestion
        if (GradleVersion.current().compareTo(GradleVersion.version("6.6")) >= 0) {
```

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
```diff
@@ -59,7 +64,19 @@ public TestFramework createRetrying(JvmTestExecutionSpec spec, Test testTask, Se
...
+            final ObjectFactory objectFactory = ((ProjectInternal) testTask.getProject()).getServices().get(ObjectFactory.class);
+            return new TestNGTestFramework(testTask, objectFactory.fileCollection(), retriedTestFilter, objectFactory);
+        } else {
+            try {
+                Class<?> testNGTestFramework = Class.forName("org.gradle.api.internal.tasks.testing.testng.TestNGTestFramework");
```
marcphilipp, line 73:
```suggestion
                Class<?> testNGTestFramework = TestNGTestFramework.class;
```

plugin/build.gradle.kts
```diff
@@ -37,7 +38,7 @@ dependencies {
     plugin("org.ow2.asm:asm:7.2")
 
     testImplementation(gradleTestKit())
-    testImplementation("org.codehaus.groovy:groovy-all:2.5.8")
+    testImplementation("org.codehaus.groovy:groovy-all:${GroovySystem.getVersion()}")
```
marcphilipp, line 41: ❓ Was the intention here to depend on a certain version of Groovy that's independent of the one used in Gradle? Otherwise `testImplementation(localGroovy())` can be used.

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
```diff
@@ -59,7 +64,19 @@ public TestFramework createRetrying(JvmTestExecutionSpec spec, Test testTask, Se
                 }
             });
 
-        return new TestNGTestFramework(testTask, retriedTestFilter, instantiator, classLoaderCache);
+        VersionNumber gradleVersion = VersionNumber.parse(testTask.getProject().getGradle().getGradleVersion());
+        if (gradleVersion.getMajor() >= 6 && gradleVersion.getMinor()>= 6) {
+            final ObjectFactory objectFactory = ((ProjectInternal) testTask.getProject()).getServices().get(ObjectFactory.class);
+            return new TestNGTestFramework(testTask, objectFactory.fileCollection(), retriedTestFilter, objectFactory);
```
marcphilipp, line 70: I wonder why this doesn't fail any test but we should pass the test classpath:

```suggestion
            return new TestNGTestFramework(testTask, testTask.getClasspath(), retriedTestFilter, objectFactory);
```

samples/sample-junit5/junit5.sample.out
marcphilipp: Hmm, that's weird indeed. I think it was introduced in 6.6. Maybe there's some special logic in Gradle for the old `PlaceholderException` that's missing for `PlaceholderAssertionError`?

sample-tests/build.gradle.kts
```diff
@@ -7,18 +10,32 @@ repositories {
...
-val copySamples by tasks.registering(Sync::class) {
-    from("../samples")
-    into("$buildDir/samples")
-}
+val snippetsDir = file("../samples")
+val processedSnippetsDir = file("$buildDir/samples")
+
+val tokens: Map<String, Provider<String>> = mapOf(
+    "test-retry-plugin-version" to provider {
+        PluginsVersionData.latestVersion("org.gradle", "test-retry-gradle-plugin")
```
marcphilipp, line 23: ❓ This should test the locally built version, shouldn't it?

marcphilipp: It seems that's currently not possible with Exemplar: https://github.com/gradle/exemplar/issues/38

samples/sample-junit5/junit5.sample.out
marcphilipp: Looks like it's caused by this:
https://github.com/gradle/gradle/blob/4cff28254160fe259563d0cffb92229746c092fc/subprojects/testing-base/src/main/java/org/gradle/api/internal/tasks/testing/logging/ShortExceptionFormatter.java#L50-L51

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
jonnybbb: Had the same in mind - but it does not work for the `rc` version

samples/sample-junit5/junit5.sample.out
marcphilipp: I've submitted https://github.com/gradle/gradle/pull/13995 to fix this in Gradle.

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
jonnybbb: This would assume that the package will not change in later versions - which I think is save.

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
marcphilipp: This should work:
```suggestion
        if (GradleVersion.current().getBaseVersion().compareTo(GradleVersion.version("6.6")) >= 0) {
```

sample-tests/build.gradle.kts
jonnybbb: I guess the intention is to test the example with the latest released public available version.

plugin/build.gradle.kts
jonnybbb: `testImplementation(localGroovy())` was the intention

plugin/src/main/java/org/gradle/testretry/internal/framework/TestNgTestFrameworkStrategy.java
jonnybbb: changed as suggested

samples/sample-junit5/junit5.sample.out
jonnybbb: will this fix result in printing `PlaceholderAssertionError` instead of `org.gradle.internal.serialize.PlaceholderAssertionError`?

marcphilipp: No, it will result in printing `org.opentest4j.AssertionFailedError`.

sample-tests/build.gradle.kts
```diff
@@ -7,18 +9,32 @@ repositories {
...
+        from(snippetsDir)
+        into(processedSnippetsDir)
+        inputs.properties(tokens)
+        doFirst {
+            filter(org.apache.tools.ant.filters.ReplaceTokens::class, "tokens" to tokens.mapValues { it.value.get() })
+        }
+    }
+    test {
+        val fileProvider = project.objects.directoryProperty().fileProvider(replaceTokensInSnippets.map { it.destinationDir })
+        inputs.files(fileProvider)
```
marcphilipp, line 38: We should preserve the `PathSensitivity` otherwise the default is `ABSOLUTE` which breaks cacheability, right?

```suggestion
        inputs.files(fileProvider)
            .withPathSensitivity(PathSensitivity.RELATIVE)
            .withPropertyName("samples")
```

sample-tests/build.gradle.kts
```diff
@@ -7,18 +9,32 @@ repositories {
...
+    val replaceTokensInSnippets by registering(Copy::class) {
+        from(snippetsDir)
+        into(processedSnippetsDir)
+        inputs.properties(tokens)
+        doFirst {
+            filter(org.apache.tools.ant.filters.ReplaceTokens::class, "tokens" to tokens.mapValues { it.value.get() })
+        }
+    }
+    test {
+        val fileProvider = project.objects.directoryProperty().fileProvider(replaceTokensInSnippets.map { it.destinationDir })
```
marcphilipp, line 37: Do we need the intermediate `DirectoryProperty` or can we simply pass `replaceTokensInSnippets.map { it.destinationDir }` to `inputs.files()`?

jonnybbb: great suggestion - fixed

sample-tests/build.gradle.kts
jonnybbb: great suggestion - fixed

sample-tests/src/test/groovy/org/gradle/testretry/samples/PlaceholderAssertionErrorOutputNormalizer.groovy
```diff
@@ -7,7 +7,7 @@ class PlaceholderAssertionErrorOutputNormalizer implements OutputNormalizer {
     @Override
     String normalize(String commandOutput, ExecutionMetadata executionMetadata) {
         def result = commandOutput.split("\\r?\\n").collect { line ->
-            line.contains("org.gradle.internal.serialize.PlaceholderAssertionError") ? line.replace("org.gradle.internal.serialize.PlaceholderAssertionError", "PlaceholderAssertionError") : line
+            line.contains('org.gradle.internal.serialize.PlaceholderAssertionError') ? line.replace('org.gradle.internal.serialize.PlaceholderAssertionError', 'org.opentest4j.AssertionFailedError') : line
```
marcphilipp, line 10: I think we should link to https://github.com/gradle/gradle/pull/13995 here so we remember why this is here... 😉

Issue #43, opened 2020-04-14
Only retry annotated tests


andrey-bolduzev:
Are there any plans to make the plugin optionally more selective in regards to which tests are being retried, for example by making the plugin config accept an annotation like `@FlakyTest` and only retrying tests annotated with it?

For instance, currently, the plugin is applied module-wide. Let's say I have some tests that are flaky for a reason outside of my control (a bug in a library the project depends on), and I would like to rerun those when they fail. At the same time, I'm afraid of adding new tests that are "flaky" due to an error in my own implementation, and I don't want new tests to be automatically retried only because they are declared in the module the plugin is applied to.

markfi-stripe:
+1 to this feature request. I have a suite of flaky by nature integration tests that I'd like to use this plugin for, however these tests are colocated with non-flaky tests in the same Gradle module.

Being able to configure the retry behavior on a per Test file basis in the source code with an annotation would be great. Even better would be if we could configure retry behavior with a method level annotation on a per test basis.

In addition to the benefit of this being more configurable, this also feels like it would cause less confusion for someone debugging a flaky test. Assuming someone starts diagnosing a flaky test, not even knowing that this plugin is being used at the Gradle scripts level, there is no indication in the test source code that there is any retrying involved -- and the first instinct is probably not to see if this Gradle plugin is installed via the adjacent build scripts.

If there were a `@FlakyTest(maxRetries = 1)` annotation on the test class or method in the source code, it would be very clear that the test is being retried.

In the meantime I will separate my flaky integration tests into a separate module and only use this plugin in that module.

asarkar:
I'd say that in addition to, or as an alternative to, annotations, it's useful to be able to target tests by glob path in the build file. That way, I can target a package or multiple classes at once without having to touch each source file.

ldaley:
The proposed implementation of this is here: https://github.com/gradle/test-retry-gradle-plugin/pull/80

It supports filtering by class name and annotation.

ldaley:
This will be in 1.10.

asarkar:
Thanks, is the documentation updated to show the usage?

ldaley:
@asarkar there is a little here: https://github.com/gradle/test-retry-gradle-plugin#filtering

Please let me know if that's inadequate.

PR #80, opened 2020-11-20
Allow specifying which classes may be retried

ldaley:
Supports class name patterns and annotation class name patterns.

Addresses https://github.com/gradle/test-retry-gradle-plugin/issues/43.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
@@ -42,7 +42,7 @@ public final class TestTaskConfigurer {
 
         TestRetryTaskExtension extension = objectFactory.newInstance(DefaultTestRetryTaskExtension.class);
 
-        TestRetryTaskExtensionAdapter adapter = new TestRetryTaskExtensionAdapter(providerFactory, extension, supportsPropertyConventions(gradleVersion));
+        TestRetryTaskExtensionAdapter adapter = new TestRetryTaskExtensionAdapter(providerFactory, extension, gradleVersion);
 
         test.getInputs().property("retry.failOnPassedAfterRetry", adapter.getFailOnPassedAfterRetryInput());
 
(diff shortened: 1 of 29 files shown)


plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -40,12 +40,7 @@ private TestTaskConfigurer() {
     public static void configureTestTask(Test test, ObjectFactory objectFactory, ProviderFactory providerFactory) {
         VersionNumber gradleVersion = VersionNumber.parse(test.getProject().getGradle().getGradleVersion());
 
-        TestRetryTaskExtension extension;
-        if (supportsGeneratedAbstractTypeImplementations(gradleVersion)) {
```
eriwen, line 44: ❓ It's not obvious to me why this check is no longer needed. What am I missing?

plugin/src/test/groovy/org/gradle/testretry/FilterFuncTest.groovy
```diff
@@ -0,0 +1,69 @@
...
+ */
+package org.gradle.testretry
+
+
+import spock.lang.Unroll
+
+class FilterFuncTest extends AbstractGeneralPluginFuncTest {
+
+    @Unroll
+    def "can filter by class include pattern (gradle version #gradleVersion)"() {
```
eriwen, line 24: I don't think I saw tests for excludes or filtering by annotation classes. I'm sure you're planning on it. Here's a reminder in case you get busy with other stuff and have to come back to this.

eriwen (commented):
In general I think this is the right UI for this kind of feature. We don't enforce that users use an annotation, but they can (use any) if they want. Alternatively it's rather easy for users to include common groups of tests like "*IntegrationTest". Configuring this in the build is important to ensure new flaky tests matching the filters are detected. 

I also like balance of applying includes and excludes, because I think users will have the flexibility to select just the set of tests they want. It might be worth some testing to demonstrate that the filters behave as we expect when there are a variety of "include/exclude" declarations.

marcphilipp (commented):
In general, I think this is a useful feature and the user-facing API looks sound. However, I'm wondering what's the use case for supporting wildcards for the annotation class names. Since we'll also have to implement this for test distribution, I'd like to keep it as simple as possible.

plugin/src/test/groovy/org/gradle/testretry/FilterFuncTest.groovy
ldaley: Indeed. I didn't add them before putting this proposal up, but they are there now.

ldaley:
>  what's the use case for supporting wildcards for the annotation class names

@marcphilipp nothing explicit or major. Two minor points in favour are symmetry with the class name control, and avoiding the need for specifying package.

In terms of implementation, supporting wildcards for this doesn't really add any complexity.

eriwen (approved):

Issue #106, opened 2021-06-24
Gradle plugin should be idempotent in the face of multiple applications from different classloaders


eriwen:
Gradle's idempotence guarantee breaks down when a plugin is applied by multiple classloaders. 

This manifests in runtime failures when an init script is used to apply this plugin, and the plugin is applied again in a project build script.

PR #107, opened 2021-06-24
Do nothing on subsequent applications of this plugins

eriwen:
This change prevents multiple applications of this plugin from
multiple classloaders from raising errors by checking whether a
plugin with the same qualified class name is already applied to the
given project, and returning early if so.

Issue: #106

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java
+++ b/plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java
@@ -39,9 +39,16 @@ public class TestRetryPlugin implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
+        if (pluginAlreadyApplied(project)) {
+            return;
+        }
+
         project.getTasks()
             .withType(Test.class)
             .configureEach(task -> configureTestTask(task, objectFactory, providerFactory));
     }
 
+    private static boolean pluginAlreadyApplied(Project project) {
+        return project.getPlugins().stream().anyMatch(plugin -> plugin.getClass().getName().equals(TestRetryPlugin.class.getName()));
+    }
 }


plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java
```diff
@@ -39,9 +39,16 @@
...
+            return;
+        }
+
         project.getTasks()
             .withType(Test.class)
             .configureEach(task -> configureTestTask(task, objectFactory, providerFactory));
     }
 
+    private static boolean pluginAlreadyApplied(Project project) {
+        return project.getPlugins().stream().anyMatch(plugin -> plugin.getClass().getName().equals(TestRetryPlugin.class.getName()));
```
jonnybbb, line 52: Why not using directly 
```suggestion
        return project.getPlugins().stream().anyMatch(plugin -> plugin.getClass().equals(TestRetryPlugin.class));
 ```

File: plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java [hunk 2]
```diff
@@ -39,9 +39,16 @@
...
+        if (pluginAlreadyApplied(project)) {
+            return;
+        }
+
         project.getTasks()
             .withType(Test.class)
             .configureEach(task -> configureTestTask(task, objectFactory, providerFactory));
     }
 
+    private static boolean pluginAlreadyApplied(Project project) {
```
jonnybbb, line 51: A matter of preference but how about 
```suggestion
    private static boolean isPluginAlreadyApplied(Project project) {
```

jonnybbb (approved):
My suggestions are not blocking this PR - only minor suggestions - otherwise 👍

plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java
```diff
@@ -39,9 +39,16 @@
 
     @Override
     public void apply(Project project) {
+        if (pluginAlreadyApplied(project)) {
```
jonnybbb, line 42:
```suggestion
        if (isPluginAlreadyApplied(project)) {
```

plugin/src/main/java/org/gradle/testretry/TestRetryPlugin.java
paplorinc: is this Config Cache safe?

eriwen: @jonnybbb We're comparing the name because the class may be coming from 2 different classloaders. 

@paplorinc This is the strategy we use for the build scan plugin, so I believe it is config cache-safe.

eriwen:
Thank you for the reviews :)

Issue #110, opened 2021-06-30
Links in README are broken


eugene-krivobokov:
What I see in README.adoc:

```txt
The plugin has been tested with the reporting of link:url[TeamCity, https://www.jetbrains.com/teamcity] and link:url[Jenkins, https://www.jenkins.io].
```

In browser, all links are resolved to `https://github.com/gradle/test-retry-gradle-plugin/blob/master/url`

paplorinc:
Thanks for reporting it @eugene-krivobokov, fixed!

PR #111, opened 2021-07-01
Fix URLs in readme

paplorinc:
Fixes https://github.com/gradle/test-retry-gradle-plugin/issues/110

Before:
Screen Shot 2021-07-01 at 9 29 14 AM

After:
Screen Shot 2021-07-01 at 9 29 26 AM


marcphilipp (approved):
Thanks! 👍

Issue #105, opened 2021-06-17
"org.gradle.test-retry was unable to retry the following test methods, which is unexpected" is thrown when second try fails with initializationError


rovner:
Gradle build log
```
./gradlew test

> Task :test FAILED

ExampleTest > test() FAILED
    java.lang.AssertionError at ExampleTest.java:29

ExampleTest > initializationError FAILED
    java.lang.RuntimeException at ExampleTest.java:22

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues
     com.example.ExampleTest#test()


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

build.gradle
```
plugins {
    id 'java'
    id "org.gradle.test-retry" version "1.2.1"
}

group 'org.example'
version '1.0-SNAPSHOT'

repositories {
    mavenCentral()
}

dependencies {
    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
}


//to simulate fail
task clarFlag() {
    doLast {
        file("build/flag").delete()
    }
}

test {
    useJUnitPlatform()
    retry {
        maxRetries = 3
        maxFailures = 20
        failOnPassedAfterRetry = true
    }
    finalizedBy clarFlag
}
```

Example test
```
package com.example;

import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.Test;

import java.io.IOException;
import java.nio.charset.StandardCharsets;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;

class ExampleTest {
    private final static Path flag = Paths.get("build/flag");

    @BeforeAll
    static void beforeAll() throws IOException {
        System.out.println("BEFORE");
        if (!flag.toFile().exists()) {
            Files.write(flag, "".getBytes(StandardCharsets.UTF_8));
        } else {
            throw new RuntimeException();
        }
    }

    @Test
    void test() {
        System.out.println("TEST");
        throw new AssertionError("TEST");
    }
}

```

marcphilipp:
[...]

PR #112, opened 2021-07-16
Add option to force entire test class to be retried

bigdaz:
In certain circumstances, tests may be structured in a way that makes them more likely to succeed when the entire test class is re-executed.  This commit adds a new option to enable this behaviour.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -146,6 +146,15 @@ public interface TestRetryTaskExtension {
     Property<Integer> getMaxFailures();
 
     /**
+     * Whether test classes should be retried in entirety, or if individual tests methods can be retried.
+     * <p>
+     * This setting defaults to {@code false}, meaning that only the test methods that actually fail will be retried.
+     *
+     * @return whether individual test methods or entire test classes should be retried
+     */
+    Property<Boolean> getRetryEntireTestClass();
+
+    /**
      * The filter for specifying which tests may be retried.
      */
     Filter getFilter();
--- a/plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
+++ b/plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
@@ -70,6 +70,15 @@ public interface TestRetryTaskExtension {
     Property<Integer> getMaxFailures();
 
     /**
+     * Whether test classes should be retried in entirety, or if individual tests methods can be retried.
+     * <p>
+     * This setting defaults to {@code false}, meaning that only the test methods that actually fail will be retried.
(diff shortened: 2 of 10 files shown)


KroArtem:
I believe it can help to reduce problems with [this issue,](https://github.com/gradle/test-retry-gradle-plugin/issues/105) right?

README.adoc
```diff
@@ -145,6 +145,15 @@ public interface TestRetryTaskExtension {
...
     Property<Integer> getMaxFailures();
 
+    /**
+     * Whether test classes should be retried in entirety, or if individual tests methods can be retried.
+     * <p>
+     * This setting defaults to {@code false}, meaning that only the test methods that actually fail will be retried.
+     *
+     * @return whether individual test methods or entire test classes should be retried
+     */
+    Property<Boolean> getRetryEntireTestClass();
```
ldaley, line 155: Are we sure we want to do it this “coarsely”? Should we instead allow classes to opt-in via an extension? I think the current proposed is OK, but wanted to put the question on the record.

KroArtem: From my point of view, when you know that this or that class/set of classes requires retries, this is indeed something that could be defined as an annotation on a class but till now there are no annotations/anything that is exposed directly in tests. 

Another approach to set a filter in gradle config would also be fine. If we relaunch everything that may lead to unnecessary overhead.

pgraves5:
FYI - We are looking for this change since we have classes of tests configured and some individual tests will not run with a retry, unless all of the previous steps (tests) in the class are executed.   We do not need to re-execute/launch ALL classes, just the classes where there was a failure from one of the individual tests.

README.adoc
bigdaz: I agree that annotating a class to be retried in entirety (or alternatively registering the class with the `retry` extension) would be a better, more fine-grained solution. We could take a similar approach to flag particular tests as needing a retry, rather than doing this in a coarse-grained manner.

I'm definitely open to this idea, but I don't think I'll have capacity to do the implementation any time soon.

bigdaz:
@pgraves5 Thanks for the feedback. The behaviour you need is indeed what is implemented in this PR. I will update the documentation/names to make it clear that we will only retry a particular test class if one or more tests for that class fail.

bigdaz:
I'm no longer convinced this is well modelled and I'm not planning on working on this further.

pbandura:
> I'm no longer convinced this is well modelled and I'm not planning on working on this further.

Sad to hear. I was really looking forward to use this feature.
Anyway, Thanks for your effort.

PR #120, opened 2021-11-16
Make build-cache-configuration config cache compatible

JLLeitschuh:


ldaley:
Thanks Jonathan.

runningcode:
Can we do this in a backward compatible with Gradle 5.x and 6.x?
Many customers are stuck on older Gradle versions but still want to benefit from the insights of test-retry functionality.

etiennestuder:
> Can we do this in a backward compatible with Gradle 5.x and 6.x? Many customers are stuck on older Gradle versions but still want to benefit from the insights of test-retry functionality.

Strong +1

etiennestuder:
Aren't there any Gradle-cross-version tests that would cache this?

JLLeitschuh:
Supposedly, according to @wolfs this is no longer required in the latest version of Gradle. These API's have been deprecated in 7.4

runningcode:
It would be good to implement this in order to get configuration cache compatibility while waiting for 7.4 but also for customers who aren't able to update to Gradle 7.4 (backward compatibility).

JLLeitschuh:
This is exclusively a settings script change. It's not part of the code shipped to end users. There is no need to merge this if this build updates straight from 7.2 to 7.4

PR #132, opened 2022-05-02
Fix breaking change introduced in Gradle 7.6

donat:
Gradle 7.6 broke the internal `TestResultProcessor` API as it changed the signature of the `failure()` method. This PR fixes the compilation error and maintains compatibility with existing Gradle versions by using reflection.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
@@ -120,12 +120,16 @@ public final class RetryTestExecuter implements TestExecuter<JvmTestExecutionSpe
     }
 
     public void failWithNonRetriedTestsIfAny() {
-        if (extension.getSimulateNotRetryableTest() || lastResult != null && !lastResult.nonRetriedTests.isEmpty()) {
+        if (extension.getSimulateNotRetryableTest() || hasNonRetriedTests()) {
             throw new IllegalStateException("org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues" +
                 lastResult.nonRetriedTests.stream()
                     .flatMap(entry -> entry.getValue().stream().map(methodName -> "   " + entry.getKey() + "#" + methodName))
                     .collect(Collectors.joining("\n", "\n", "\n")));
         }
+    }
+
+    private boolean hasNonRetriedTests() {
+        return lastResult != null && !lastResult.nonRetriedTests.isEmpty();
     }
 
     private JvmTestExecutionSpec createRetryJvmExecutionSpec(JvmTestExecutionSpec spec, TestFramework retryTestFramework) {
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
@@ -120,7 +120,7 @@ public final class RetryTestExecuter implements TestExecuter<JvmTestExecutionSpe
     }
 
     public void failWithNonRetriedTestsIfAny() {
-        if (extension.getSimulateNotRetryableTest() || (lastResult != null && !lastResult.nonRetriedTests.isEmpty())) {
+        if (extension.getSimulateNotRetryableTest() || lastResult != null && !lastResult.nonRetriedTests.isEmpty()) {
             throw new IllegalStateException("org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues" +
                 lastResult.nonRetriedTests.stream()
                     .flatMap(entry -> entry.getValue().stream().map(methodName -> "   " + entry.getKey() + "#" + methodName))
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -171,7 +171,11 @@ final class RetryTestResultProcessor implements TestResultProcessor {
     private boolean lastRun() {
         return currentRoundFailedTests.isEmpty()
             || lastRetry
-            || maxFailures > 0 && currentRoundFailedTests.size() >= maxFailures;
+            || currentRoundFailedTestsExceedsMaxFailures();
+    }
+
+    private boolean currentRoundFailedTestsExceedsMaxFailures() {
+        return maxFailures > 0 && currentRoundFailedTests.size() >= maxFailures;
     }
 
     public RoundResult getResult() {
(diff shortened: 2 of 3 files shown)


sorin-florea (approved):

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -124,8 +127,27 @@ public void output(Object testId, TestOutputEvent testOutputEvent) {
...
+    @SuppressWarnings("unused")
     public void failure(Object testId, Throwable throwable) {
+        // Gradle 7.6 changed the method signature from failure(Object, Throwable) to failure(Object, TestFailure).
+        // To maintain compatibility with older versions, the original method needs to exist and needs to call failure()
+        // on the delegate via reflection.
+        failure(testId);
+        try {
+            Method failureMethod = delegate.getClass().getMethod("failure", Object.class, Throwable.class);
+            failureMethod.invoke(delegate, testId, throwable);
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
```
marcphilipp, line 139:
```suggestion
        } catch (ReflectiveOperationException e) {
```

plus import

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -124,8 +127,27 @@ public void output(Object testId, TestOutputEvent testOutputEvent) {
...
 
-    @Override
+    @SuppressWarnings("unused")
     public void failure(Object testId, Throwable throwable) {
+        // Gradle 7.6 changed the method signature from failure(Object, Throwable) to failure(Object, TestFailure).
+        // To maintain compatibility with older versions, the original method needs to exist and needs to call failure()
+        // on the delegate via reflection.
+        failure(testId);
+        try {
+            Method failureMethod = delegate.getClass().getMethod("failure", Object.class, Throwable.class);
```
marcphilipp, line 137: ❓ Should we cache the resulting method instead of performing this lookup each time?

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -124,8 +127,27 @@ public void output(Object testId, TestOutputEvent testOutputEvent) {
...
+            failureMethod.invoke(delegate, testId, throwable);
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void failure(Object testId, TestFailure result) {
```
marcphilipp, line 145: ❓ Won't this cause problems when loading this class on previous Gradle versions?

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
marcphilipp: ❌ To answer my own question: I think we should. For simplicity we can look it up in the constructor and cache it as an instance variable.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
marcphilipp: According to our cross-version tests, this works: https://e.grdev.net/s/eoxryno34i4n6

marcphilipp (changes requested):

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
donat: Done

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
donat: Sorry for not responding earlier; will do that.

donat: The method will method only exists for Gradle versions <7.6, so we cannot do the lookup in the constructor (unless can investigate the Gradle version in this class, but that seems like an overkill). A mutable variable cache should suffice.

donat: Done

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
donat: That was my assumption, the TestFailure class should be loaded just-in-time, before the `failure(Object, TestFailure)` method is called.

donat:
@marcphilipp ptal.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
```diff
@@ -120,7 +120,7 @@ public void execute(JvmTestExecutionSpec spec, TestResultProcessor testResultPro
     }
 
     public void failWithNonRetriedTestsIfAny() {
-        if (extension.getSimulateNotRetryableTest() || (lastResult != null && !lastResult.nonRetriedTests.isEmpty())) {
+        if (extension.getSimulateNotRetryableTest() || lastResult != null && !lastResult.nonRetriedTests.isEmpty()) {
```
marcphilipp, line 123: Was there an IDEA warning for these? I think we should instead move the `&&` expression to keep this readable.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -164,7 +171,7 @@ private void failure(Object testId) {
     private boolean lastRun() {
         return currentRoundFailedTests.isEmpty()
             || lastRetry
-            || (maxFailures > 0 && currentRoundFailedTests.size() >= maxFailures);
+            || maxFailures > 0 && currentRoundFailedTests.size() >= maxFailures;
```
marcphilipp, line 174: Ditto.

marcphilipp (changes requested):
Looks good except for two minor readability issues.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestExecuter.java
donat: The checkstyle task fails without the change: https://e.grdev.net/s/7wkocu7ovlrhc/failure#1.

marcphilipp: I guess that's because the default `toolVersion` changed recently: https://github.com/gradle/gradle/commit/b4d55d51d7cc9ae971ef47b0ff66b94c7a365ea7

marcphilipp: Done in 9d3cbe71ea7ed1e4aff5b685d2b96caf3a97cea5.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
marcphilipp: Done in 9d3cbe71ea7ed1e4aff5b685d2b96caf3a97cea5.

marcphilipp (approved):

PR #134, opened 2022-05-11
Exclude spot instances from CI

sorin-florea:


marcphilipp:
@sorin-florea What triggered this change?

sorin-florea:
@marcphilipp This was a request from @jonnybbb to exclude spot instances since those can be terminated at any point.

jonnybbb (changes requested):
Only release related jobs should **not** run on EC2 agents. Please have a look into https://github.com/gradle/test-retry-gradle-plugin/blob/master/.teamcity/settings.kts - I might have guided you wrong in the first place - sorry.

PR #138, opened 2022-06-29
Retry on TestNG BeforeMethod/AfterMethod failure

anaberezhnov:
Fix "org.gradle.test-retry was unable to retry the following test methods" that are annotated with TestNG BeforeMethod/AfterMethod by retrying the entire class. The entire class is retried because TestNG skips tests on lifecycle method failure.


marcphilipp (approved):
Thanks for the PR! 👍

Issue #101, opened 2021-05-10
retry does not work with JUnitParamsRunner


sereneliu:
I get this error message for tests with JUnitParamsRunner:

`org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues`

What can I tweak to make this work?

daknin:
I've also encountered this. It looks like the `JUnitParamsRunner` sets the test names to be something like `parameterizedMethod(1) [0]` which `BaseJunitTestFrameworkStrategy::addPotentiallyParameterizedSuffixed` incorrectly changes to  `parameterizedMethod(1) ` instead of `parameterizedMethod`

Changing `String strippedParameterName = name.replaceAll("(?:\\([^)]*?\\)|\\[[^]]*?])*$", "");` to `String strippedParameterName = name.replaceAll("(?:\\s?\\([^)]*?\\)|\\s?\\[[^]]*?])*$", "");` seems to fix it

Sample test
```java
package org.example;

import junitparams.JUnitParamsRunner;
import junitparams.Parameters;
import junitparams.naming.TestCaseName;
import org.junit.Test;
import org.junit.experimental.categories.Category;
import org.junit.runner.RunWith;

import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.*;

@RunWith(JUnitParamsRunner.class)
public class MainTest {
	@Test
	@Parameters({
			"17, true",
			"22, true" })
	@TestCaseName("parameterized{0}_{1}")
	public void parameterizedMethod(int value, boolean valid) {
		assertThat(value > 30, is(valid));
	}
}
```

build.gradle
```groovy
plugins {
    id 'java'
	id "org.gradle.test-retry" version "1.4.0"
}

group 'org.example'
version '1.0-SNAPSHOT'

repositories {
    mavenCentral()
}

dependencies {
	testImplementation 'junit:junit:4.13.2'
	implementation 'pl.pragmatists:JUnitParams:1.1.1'
	testImplementation 'org.junit.vintage:junit-vintage-engine:5.9.0'

}

test {
	retry {
		maxRetries.set(2)
		maxFailures.set(20)
		failOnPassedAfterRetry.set(true)
	}
    useJUnit() {
	}
}

```

marcphilipp:
[...]

PR #143, opened 2022-08-16
#101 - Adds support for JUnitParams / JUnit4

daknin:
As described in #101, using the plugin with [JUnitParams](https://github.com/Pragmatists/JUnitParams) results in a failure message. 

`org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues`

This PR updates the regex used to extract the Java method name from the JUnit test name. It also adds a test / example for using JUnitParams tests with the plugin


marcphilipp:
@daknin Could you please rebase this on `main` to fix the Gradle build? The snapshot version of Gradle used on this branch is no longer available.

daknin:
@marcphilipp I've rebased as requsted

marcphilipp:
@daknin The new test fails for Gradle 5.0 - 5.6.4. Could you please take a look?

PR #146, opened 2022-08-29
Update readme with workaround of unsupported test tasks

cdsap:
Update readme with workaround of unsupported test tasks


marcphilipp (changes requested):
I think we should go with #147 instead.

Issue #148, opened 2022-09-19
Compatibility with Gradle 8.0


marcphilipp:
## Motivation

A recent change to internal APIs in Gradle (https://github.com/gradle/gradle/pull/21919) broke compatibility.

A potential solution would be to add a `copy(TestFilter): TestFramework` method to `TestFramework`. Another workaround would be to use reflection to call the old or new constructor.

https://github.com/gradle/test-retry-gradle-plugin/blob/fd1933e424de5d79c1c5c1d18e31896fc0c90380/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/JunitTestFrameworkStrategy.java#L27-L34

snoopcheri:
I had a quick call with @jvandort :
* a copy method on the `TestFramework` level makes the most sense
* we have to see, whether we can define this method on the interface level (depends on required arguments for the specific test frameworks)
* adding `@UsedByScanPlugin("test-retry")` on this additional method is also a very good idea
* @snoopcheri will create a PR and add Justin, Sterling and Marc as reviewers - Justin will possibly add the whole core-team as reviewers

snoopcheri:
Draft PR in gradle/gradle: https://github.com/gradle/gradle/pull/22139
Draft PR in gradle/test-retry-gradle-plugin: https://github.com/gradle/test-retry-gradle-plugin/pull/149

PR #149, opened 2022-09-22
Add support for Gradle 8

snoopcheri:
This PR adds compatibility with the recent changes in Gradle 8 nightly.

It is based on the changes in https://github.com/gradle/gradle/pull/22139 and shows how those changes would be used in the test retry plugin.

Relates to https://github.com/gradle/test-retry-gradle-plugin/issues/148


plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
```diff
@@ -16,24 +16,55 @@
...
+    }
+
+    private static JUnitPlatformTestFramework newJUnitPlatformTestFrameworkInstanceForGradleOlderThanV8_0(Test task, DefaultTestFilter failedTestsFilter) {
+        try {
+            Class<?> jUnitPlatformTestFrameworkClass = JUnitPlatformTestFramework.class;
+            Constructor<?> constructor = jUnitPlatformTestFrameworkClass.getConstructor(DefaultTestFilter.class);
+
+            return (JUnitPlatformTestFramework) constructor.newInstance(failedTestsFilter);
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException |
+                 InvocationTargetException e) {
```
marcphilipp, line 62:
```suggestion
        } catch (ReflectiveOperationException e) {
```

marcphilipp (approved):

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
snoopcheri: DONE (26701fe8)

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
```diff
@@ -16,27 +16,82 @@
...
+            }
+
+            private static JUnitPlatformTestFramework newInstance(DefaultTestFilter failedTestsFilter) {
+                try {
+                    Class<?> jUnitPlatformTestFrameworkClass = JUnitPlatformTestFramework.class;
+                    Constructor<?> constructor = jUnitPlatformTestFrameworkClass.getConstructor(DefaultTestFilter.class);
+
+                    return (JUnitPlatformTestFramework) constructor.newInstance(failedTestsFilter);
+                } catch (ReflectiveOperationException e) {
+                    throw new RuntimeException(e);
```
leonard84, line 75: Don't we have `UncheckedException` for this, or some other more fitting type?

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
```diff
@@ -16,27 +16,82 @@
...
+            private final TestFramework testFramework;
+
+            ProviderForCurrentGradleVersion(TestFramework testFramework) {
+                this.testFramework = testFramework;
+            }
+
+            @Override
+            public TestFramework testFrameworkFor(DefaultTestFilter failedTestsFilter) {
+                return testFramework.copyWithFilters(failedTestsFilter);            }
+        }
```
leonard84, line 50: ❓ Why can't we share this implementation, it is the same across all frameworks.

leonard84 (changes requested):

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
marcphilipp: I was also wondering that.

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
```diff
@@ -16,27 +16,82 @@
...
+
+            private final TestFramework testFramework;
+
+            ProviderForCurrentGradleVersion(TestFramework testFramework) {
+                this.testFramework = testFramework;
+            }
+
+            @Override
+            public TestFramework testFrameworkFor(DefaultTestFilter failedTestsFilter) {
+                return testFramework.copyWithFilters(failedTestsFilter);            }
```
marcphilipp, line 49:
```suggestion
                return testFramework.copyWithFilters(failedTestsFilter);
            }
```

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
snoopcheri: I've used the same exception as we're currently throwing in similar situations like the [RetryTestResultProcessor](https://github.com/gradle/test-retry-gradle-plugin/blob/main/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java#L139-L141).

I think it makes sense to always throw the same here and therefore I suggest to keep it that way in order to minimize risk that we're breaking something else with customers. Personally, I probably would have chosen `UnsupportedOperationException`.

@leonard84 WDYT?

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
snoopcheri: Good idea.

DONE (45ed11e2)

jvandort (approved):

leonard84 (approved):

PR #151, opened 2022-10-17
Decouple retry extension

snoopcheri:
This PR introduces a new interface for accessing properties on the `retry` extension. The `RetryTestExecuter` uses this interface now instead of `TestRetryTaskExtensionAdapter`.

Also, the usage of `org.gradle.util.VersionNumber` is removed, because it is deprecated in Gradle.

### DIFF
--- a/buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
+++ b/buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
@@ -1,9 +1,13 @@
 package org.gradle.testretry.build
 
 import groovy.json.JsonSlurper
-import org.gradle.util.VersionNumber
+import org.gradle.util.GradleVersion
+
+import java.util.regex.Pattern
 
 class GradleVersionData {
+
+    static final Pattern MAJOR_AND_MINOR_VERSION_PATTERN = Pattern.compile("(\\d+)\\.(\\d+).*")
 
     static List<String> getNightlyVersions() {
         def releaseNightly = getLatestReleaseNightly()
@@ -19,20 +23,43 @@ class GradleVersionData {
     }
 
     static List<String> getReleasedVersions() {
+        def GRADLE_5 = GradleVersion.version("5.0")
+        GRADLE_5.version
+
         new JsonSlurper().parse(new URL("https://services.gradle.org/versions/all"))
             .findAll { !it.nightly && !it.snapshot } // filter out snapshots and nightlies
             .findAll { !it.rcFor || it.activeRc } // filter out inactive rcs
             .findAll { !it.milestoneFor } // filter out milestones
-            .<String, VersionNumber, String>collectEntries { [(it.version): VersionNumber.parse(it.version as String)] }
-            .findAll { it.value.major >= 5 } // only 5.0 and above
-            .inject([] as List<Map.Entry<String, VersionNumber>>) { releasesToTest, version -> // only test against latest patch versions
-                if (!releasesToTest.any { it.value.major == version.value.major && it.value.minor == version.value.minor }) {
+            .<String, GradleVersion, String>collectEntries { [(it.version): GradleVersion.version(it.version as String)] }
+            .findAll { it.value >= GRADLE_5 } // only 5.0 and above
+            .inject([] as List<Map.Entry<String, GradleVersion>>) { releasesToTest, version -> // only test against latest patch versions
+                if (!releasesToTest.any { major(it.value) == major(version.value) && minor(it.value) == minor(version.value) }) {
                     releasesToTest + version
                 } else {
                     releasesToTest
                 }
             }
             .collect { it.key.toString() }
+    }
+
+    static int major(GradleVersion gradle) {
+        def version = gradle.version
+        def matcher = MAJOR_AND_MINOR_VERSION_PATTERN.matcher(version)
+        if (matcher.matches()) {
+            return Integer.parseInt(matcher.group(1))
+        } else {
+            throw new IllegalArgumentException("Failed to determine major version for version ${version}")
+        }
+    }
+
+    static int minor(GradleVersion gradle) {
+        def version = gradle.version
+        def matcher = MAJOR_AND_MINOR_VERSION_PATTERN.matcher(version)
+        if (matcher.matches()) {
+            return Integer.parseInt(matcher.group(2))
+        } else {
+            throw new IllegalArgumentException("Failed to determine major version for version ${version}")
+        }
     }
 
 }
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/gradle/testretry/internal/config/TestRetryTaskExtensionAccessor.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.testretry.internal.config;
+
+import java.util.Set;
+
+public interface TestRetryTaskExtensionAccessor {
```
snoopcheri, line 20: I'm not sure whether this should outside the internal package, as we will use this interface.

At the moment, we only have:
* `TestRetryPlugin`
* `TestRetryTaskExtension`

on the top level. All the rest is inside an `internal` package.

Should the read-only interface for the extension also be outside?WDYT?

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -35,4 +42,24 @@ class GradleVersionData {
             .collect { it.key.toString() }
     }
 
+    static int major(GradleVersion gradle) {
```
snoopcheri, line 45: The `GradleVersion` class unfortunately doesn't provide methods for accessing the major/minor number of the Gradle version.

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -1,10 +1,14 @@
...
 
 import groovy.json.JsonSlurper
-import org.gradle.util.VersionNumber
+import org.gradle.util.GradleVersion
+
+import java.util.regex.Pattern
 
 class GradleVersionData {
 
+    static final Pattern MAJOR_AND_MINOR_VERSION_PATTERN = Pattern.compile("(\\d+)\\.(\\d+).*")
```
leonard84, line 10: 🧐 Use groovy regex syntax, i.e., `~/regex/ and named groups
```suggestion
    static final Pattern MAJOR_AND_MINOR_VERSION_PATTERN = ~/(?<major>\d+)\.(?<minor>\d+).*/
```

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -1,10 +1,14 @@
 package org.gradle.testretry.build
 
 import groovy.json.JsonSlurper
-import org.gradle.util.VersionNumber
+import org.gradle.util.GradleVersion
+
+import java.util.regex.Pattern
 
 class GradleVersionData {
```
leonard84, line 8: 🤔 add `@CompileStatic`

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -19,14 +23,17 @@ class GradleVersionData {
     }
 
     static List<String> getReleasedVersions() {
+        def GRADLE_5 = GradleVersion.version("5.0")
+        GRADLE_5.version
```
pshevche, line 27: ❔ What does this line do? Just looks unusual :)

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -35,4 +42,24 @@ class GradleVersionData {
...
+        }
+    }
+
+    static int minor(GradleVersion gradle) {
+        def version = gradle.version
+        def matcher = MAJOR_AND_MINOR_VERSION_PATTERN.matcher(version)
+        if (matcher.matches()) {
+            return Integer.parseInt(matcher.group(2))
+        } else {
+            throw new IllegalArgumentException("Failed to determine major version for version ${version}")
```
pshevche, line 61:
```suggestion
            throw new IllegalArgumentException("Failed to determine minor version for version ${version}")
```

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
```diff
@@ -35,4 +42,24 @@ class GradleVersionData {
...
+        def matcher = MAJOR_AND_MINOR_VERSION_PATTERN.matcher(version)
+        if (matcher.matches()) {
+            return Integer.parseInt(matcher.group(1))
+        } else {
+            throw new IllegalArgumentException("Failed to determine major version for version ${version}")
+        }
+    }
+
+    static int minor(GradleVersion gradle) {
+        def version = gradle.version
```
pshevche, line 56: 🧐 Maybe `major` and `minor` methods could be collapsed? Or the common part is extracted? Both of them do the matching, just the part of the exception message and the group number are different.

plugin/src/main/java/org/gradle/testretry/internal/config/TestRetryTaskExtensionAccessor.java
pshevche: I don't mind having it inside the `internal` package. I guess we do not expect customers to interact with it, do we?

pshevche (approved):

plugin/src/main/java/org/gradle/testretry/internal/config/TestRetryTaskExtensionAccessor.java
snoopcheri: No, we don't expect customers to use it. The decoupling is really just meant for us, so we can use the `TestRetryExecuter` in Gradle Enterprise with a different implementation of the `TestRetryTaskExtension`.

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
snoopcheri: DONE (c201b530)

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
snoopcheri: DONE (032abd7b)

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
snoopcheri: FIXED via c201b530

buildSrc/src/main/groovy/org/gradle/testretry/build/GradleVersionData.groovy
snoopcheri: Thought about it but first decided against it, since it's such a trivial function. But given that it had a copy/paste error, I changed it. :P

DONE (c201b530)

marcphilipp (approved):

PR #156, opened 2022-11-10
Adjustment for creating `JvmTestExecutionSpec` when retrying tests and using Gradle 8

snoopcheri:
This PR uses the new `copyWithTestFramework` method when creating a new `JvmTestExecutionSpec` in the test retry executer.

This ensures that we do not have to adapt again, if the execution spec gets changed again in the future.

The constructor for older Gradle versions is called via reflection. Therefore, [constructors for older test-retry versions](https://github.com/gradle/gradle/blob/master/subprojects/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/JvmTestExecutionSpec.java#L45-L54) can be removed from the Gradle 8 code base.


snoopcheri:
PR: https://github.com/gradle/test-retry-gradle-plugin/pull/156

jvandort (approved):
LGTM

marcphilipp (approved):

PR #122, opened 2021-11-22
Add retryOnClassLevel option

U1F984:
I have tests which need to be retried on class level (integration tests, with individual steps being ordered using `@Order`). This PR adds a new `retryOnClassLevel` property that always retries the whole test class.

PR #157, opened 2022-11-17
Add support for retrying whole class retry

leonard84:
supersedes #122


plugin/src/main/java/org/gradle/testretry/internal/filter/ClassRetryMatcher.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public boolean retryWholeClass(String className) {
+        if (anyMatch(includeClasses, className)) {
+            return true;
+        }
+
+        Set<String> annotations; // fetching annotations is expensive, don't do it unnecessarily.
+        if (!includeAnnotationClasses.isEmpty()) {
+            annotations = annotationInspector.getClassAnnotations(className);
+            return !annotations.isEmpty() && anyMatch(includeAnnotationClasses, annotations);
```
snoopcheri, line 48: The `!annotations.isEmpty()` could be skipped.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -90,7 +94,11 @@ public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
 
                 boolean failedInPreviousRound = previousRoundFailedTests.remove(className, name);
                 if (failedInPreviousRound && testCompleteEvent.getResultType() == SKIPPED) {
-                    currentRoundFailedTests.add(className, name);
+                    if (classRetryMatcher.retryWholeClass(className)) {
```
snoopcheri, line 97: :thinking Wrap this part in private method so it can be reused from below as well?

plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
```diff
@@ -81,63 +81,107 @@ public interface TestRetryTaskExtension {
...
+     */
+    void classRetry(Action<? super ClassRetryCriteria> action);
+
+    /**
+     * A set of criteria specifying which test classes must be retried as a whole unit.
+     */
+    interface ClassRetryCriteria {
+
+        /**
+         * The patterns used to include tests based on their class name.
```
snoopcheri, line 160: This JavaDoc has to be adjusted.

We should probably also make clear that those criteria are only used for tests that pass the filter? You cannot add additional test classes for retry here.

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -136,6 +137,10 @@ abstract class AbstractPluginFuncTest extends Specification {
         testProjectDir.newFile("$sourceFilePackage/${className}.${testLanguage()}") << source
     }
 
+    void writeJavaTestSource(@Language("JAVA") String source) {
```
snoopcheri, line 140: TIL

snoopcheri (changes requested):

snoopcheri:
Once this is merged, we should publish it to our artifactory and update the version (and the copied classes) in GE as well.

plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
marcphilipp: I agree that this should be clarified but would do so in the class-level Javadoc.

plugin/src/test/groovy/org/gradle/testretry/TestDistributionIntegrationFuncTest.groovy
```diff
@@ -157,7 +157,9 @@ class TestDistributionIntegrationFuncTest extends AbstractGeneralPluginFuncTest
 
     def assertRetries(BuildResult result, int retries) {
         // 1 initial + retries + 1 overall task FAILED + 1 build FAILED
-        assert result.output.count('FAILED') == 1 + retries + 1 + 1
+        with(result.output) {
+    it.count('FAILED') == 1 + retries + 1 + 1
+}
```
marcphilipp, line 162:
```suggestion
            it.count('FAILED') == 1 + retries + 1 + 1
        }
```

plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit5FuncTest.groovy
```diff
@@ -58,37 +58,49 @@ class JUnit5FuncTest extends AbstractFrameworkFuncTest {
...
-                assert result.output.count("${beforeClassErrorTestMethodName(gradleVersion)} FAILED") == 3
-                assert result.output.count("${beforeClassErrorTestMethodName(gradleVersion)} PASSED") == 0
-                assert result.output.count('successTest() FAILED') == 0
-                assert result.output.count('successTest() PASSED') == 0
+                with(result.output) {
+    it.count("${beforeClassErrorTestMethodName(gradleVersion)} FAILED") == 3
+    it.count("${beforeClassErrorTestMethodName(gradleVersion)} PASSED") == 0
+    it.count('successTest() FAILED') == 0
+    it.count('successTest() PASSED') == 0
+}
```
marcphilipp, line 66: Indentation is off here and below.

marcphilipp (approved):
Thanks for cleaning up the test code! 👍

plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
leonard84: Well, the pattern includes these classes to be retried as a whole, so I don't think it is wrong.

> I agree that this should be clarified but would do so in the class-level Javadoc.

I don't quite understand the interface and the getter/action method are documented accordingly.

Do you have any suggestions to improve this?

plugin/src/main/java/org/gradle/testretry/internal/filter/ClassRetryMatcher.java
leonard84: It could be, we are avoiding creating a singleton list and some stream initialization with this check.

plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
```diff
@@ -81,63 +81,107 @@ public interface TestRetryTaskExtension {
...
+         * <p>
          * If no patterns are specified, all classes (that also meet other configured filters) will be included.
          */
         SetProperty<String> getExcludeAnnotationClasses();
 
     }
 
+
+    /**
+     * The set of criteria specifying which test classes must be retried as a whole unit.
```
marcphilipp, line 145:
```suggestion
     * The set of criteria specifying which test classes must be retried as a whole unit
     * if retries are enabled and the test class passes the configured {@linkplain TestRetryTaskExtension#getFilter filter}.
```

plugin/src/main/java/org/gradle/testretry/TestRetryTaskExtension.java
marcphilipp: I added a proposal above.

snoopcheri (approved):

PR #158, opened 2022-11-18
Provide better error message if `retry` extension is already present

snoopcheri:
Before registering the `retry` extension, the Test Retry Gradle plugin now checks for the presence of `retry` extension.

* If there already one registered, it will provide a proper error message.
* In case the extension uses the `com.gradle.enterprise` package, it is assumed, that the Gradle Enterprise Gradle plugin has added it. In this case, the error message includes a way how to disable the retry application by the Gradle Enterprise Gradle plugin

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
@@ -82,7 +82,7 @@ public final class TestTaskConfigurer {
                 throw new IllegalStateException(""
                     + "Another plugin is conflicting with the Test Retry Gradle plugin "
                     + "and has already added a retry extension to the test task " + testTask.getName() + ". "
-                    + "Please either remove the conflicting plugin or the Test Retry Gradle plugin from this test task."
+                    + "Please either remove the conflicting plugin or the Test Retry Gradle plugin from this test project."
                 );
             }
         }
--- a/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
@@ -75,7 +75,7 @@ public final class TestTaskConfigurer {
                     + "The Gradle Enterprise Gradle plugin is conflicting with the Test Retry Gradle plugin "
                     + "and has already added a retry extension to the test task " + testTask.getName() + ". "
                     + "Please either remove the Test Retry Gradle plugin from this project "
-                    + "or disable the application of the retry extension in the Gradle Enterprise Gradle plugin "
+                    + "or disable the registration of the retry extension in the Gradle Enterprise Gradle plugin "
                     + "by specifying the system property 'gradle.enterprise.testretry.disabled'."
                 );
             } else {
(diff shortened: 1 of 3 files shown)


pshevche (approved):

jean-andre-gauthier (approved):

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -61,6 +64,30 @@ public static void configureTestTask(Test test, ObjectFactory objectFactory, Pro
...
+                    + "and has already added a retry extension to the test task " + testTask.getName() + ". "
+                    + "Please either remove the Test Retry Gradle plugin from this test task "
+                    + "or disable the application of the retry extension in the Gradle Enterprise Gradle plugin "
+                    + "by specifying the system property 'gradle.enterprise.testretry.disabled'."
+                );
+            } else {
+                throw new IllegalStateException(""
+                    + "Another plugin is conflicting with the Test Retry Gradle plugin "
+                    + "and has already added a retry extension to the test task " + testTask.getName() + ". "
+                    + "Please either remove the conflicting plugin or the Test Retry Gradle plugin from this test task."
```
marcphilipp, line 85:
```suggestion
                    + "Please either remove the conflicting plugin or the Test Retry Gradle plugin from this test project."
```

marcphilipp (approved):

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -61,6 +64,30 @@ public static void configureTestTask(Test test, ObjectFactory objectFactory, Pro
...
+
+            if (retryExtensionClassName.startsWith(GRADLE_ENTERPRISE_BASE_PACKAGE)) {
+                throw new IllegalStateException(""
+                    + "The Gradle Enterprise Gradle plugin is conflicting with the Test Retry Gradle plugin "
+                    + "and has already added a retry extension to the test task " + testTask.getName() + ". "
+                    + "Please either remove the Test Retry Gradle plugin from this project "
+                    + "or disable the registration of the retry extension in the Gradle Enterprise Gradle plugin "
+                    + "by specifying the system property 'gradle.enterprise.testretry.disabled'."
+                );
+            } else {
```
leonard84, line 81: ❔ do we prevent double application of the retry plugin itself already elsewhere?

leonard84 (approved):

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -61,6 +64,30 @@ public static void configureTestTask(Test test, ObjectFactory objectFactory, Pro
         test.doLast(new ConditionalTaskAction(isDeactivatedByTestDistributionPlugin, new FinalizeTaskAction()));
     }
 
+    private static void ensureThatNoRetryExtensionIsPresent(Test testTask) {
+        Object existingRetryExtension = testTask.getExtensions().findByName(TestRetryTaskExtension.NAME);
```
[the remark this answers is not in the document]
snoopcheri: I've tried this with the latest version of our GE Gradle plugin (obfuscated version) and the current version of the Test-Retry Gradle plugin.

The class name of the returned extension object is `com.gradle.enterprise.gradleplugin.testretry.a.a_Decorated`.

So yes, it is a decorated version by Gradle and it keeps its package name.

Issue #140, opened 2022-07-25
`Unexpected test executer: KotlinJvmTest`


JavierSegoviaCordoba:
In a Kotlin Multiplatform project, I have tried to use this plugin and I am getting this issue, not sure if I should report this here or in the Kotlin repo.

```
> Unexpected test executer: org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest$Executor@5fbdf032
```

The config:

```kotlin
tasks.withType<Test>().configureEach {
    retry {
        maxRetries.set(5)
        maxFailures.set(100)
        failOnPassedAfterRetry.set(false)
    }
    maxParallelForks = 1
}
```

Curiously, all tests are passing and the build fail comes after that.

JavierSegoviaCordoba:
Sorry, looks like it is already reported: https://github.com/gradle/test-retry-gradle-plugin/issues/140

Issue #116, opened 2021-09-28
Cannot use this plugin in tandem with Kotlin Multiplatform Tests


eriwen:
> We got a report from one of our users that was having trouble running the test retry plugin with a kotlin project. After some digging, we found that `org.jetbrains.kotlin.multiplatform.gradle.plugin`  creates their own test executer for `KotlinJvmTest` per https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plug[…]/org/jetbrains/kotlin/gradle/targets/jvm/tasks/KotlinJvmTest.kt
When this happens, this part of the test retry plugin is reached: https://github.com/gradle/test-retry-gradle-plugin/blob/v1.3.1/plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java#L142 so tests fail with the following:

```
> Configure project :
Kotlin Multiplatform Projects are an Alpha feature. See: https://kotlinlang.org/docs/reference/evolution/components-stability.html. To hide this message, add 'kotlin.mpp.stability.nowarn=true' to the Gradle properties.


> Task :jvmTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':jvmTest'.
> Unexpected test executer: org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest$Executor@1073de17
```

[Bug reproducer project](https://github.com/nebula-plugins/gradle-nebula-integration/tree/master/kotlin-test-retry-issue)

osipxd:
Found a workaround:
```kotlin
tasks.withType<KotlinJvmTest>().configureEach {
    applyTestRetryCompatibilityWorkaround()
}

/**
 * Test retry plugin is incompatible with test tasks that override the `createTestExecuter` method.
 * This includes the [KotlinJvmTest] task which wraps the test executor with its own wrapper.
 *
 * This workaround heavily relies on the internal implementation details of the test-retry plugin and KGP.
 *
 * The test retry plugin adds a `doFirst` action, which:
 * - Retrieves the test executer using `createTestExecuter` (KGP returns wrapped test executer here)
 * - Wraps it with `RetryTestExecuter`
 * - Sets the executer using `setTestExecuter`
 *
 * In the `doLast` action, it expects that `createTestExecuter` returns the previously created `RetryTestExecuter` instance.
[...]

PR #163, opened 2022-12-07
Fix for case of inherited from Test tasks with own executers

zhurs:
Now plugin relies on task executer class, but this approach don't work with inherited tasks with own executers, for example with Kotlin Multiplatform Tests.

There are some issues for this problem:
https://github.com/gradle/test-retry-gradle-plugin/issues/116
https://github.com/gradle/test-retry-gradle-plugin/issues/140

Possible solution is to store last result outside of the retry executer.


marcphilipp:
@zhurs Could you please elaborate how this fixes the integration issue with KMPP? As far as I can tell, there would be multiple instances of the KMPP `TestExecuter` implementation created so the structure would be `KotlinJvmTest$Executor(RetryTestExecutor(KotlinJvmTest$Executor(...)))`, wouldn't it?

zhurs:
@marcphilipp You are absolutely right.
But KMMP executer just add "[jvm]" suffix to the test names, after fix suffix will be added twice - it is not very good, but at least retry will work correctly.
I don't see better way to fix issue ;-(

On the other hands fix itself is not bad - it just remove requirement of specific executer class, it will work with other Executer delegate-chains.

JavierSegoviaCordoba:
Can this be revised?

dtretyakov:
@marcphilipp do you have a better solution in mind to address the original problem with test runner in Kotlin MPP projects?

If so could you please share the plan when they will be properly supported.

pshevche:
> @marcphilipp do you have a better solution in mind to address the original problem with test runner in Kotlin MPP projects?
> 
> If so could you please share the plan when they will be properly supported.

We don't have a clear roadmap for it yet. This is a common request (also see https://github.com/gradle/test-retry-gradle-plugin/issues/116), so we'll take it into account when planning the next set of features for this plugin. Unfortunately, I don't have more to share at the moment.

PR #176, opened 2023-02-23
Handle flaky setup that fails after success and prevents failed methods from being re-tried

pshevche:
### Summary

In case of a setup that fails after successful execution, we won't see descriptors for methods from the previous run and will fail with `org.gradle.test-retry was unable to retry the following test methods, which is unexpected`. This PR solves this case by adding all failed methods from the previous run to be re-tried again.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -97,9 +97,18 @@ final class RetryTestResultProcessor implements TestResultProcessor {
                     addRetry(className, name);
                 }
 
+                // test class lifecycle which was successful but failed on retry can prevent tests from being re-tried
+                // in this case, retry all non-retried tests from previous run
+                if (isLifecycleFailure(className, name) && !failedInPreviousRound) {
+                    previousRoundFailedTests.remove(className, n -> {
+                        addRetry(className, n);
+                        return true;
+                    });
+                }
+
                 if (isClassDescriptor(descriptor)) {
                     previousRoundFailedTests.remove(className, n -> {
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -97,9 +97,18 @@ public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
                     addRetry(className, name);
                 }
 
+                // test class lifecycle which was successful but failed on retry can prevent tests from being re-tried
+                // in this case, retry all non-retried tests from previous run
+                if (isLifecycleFailure(className, name) && !failedInPreviousRound) {
+                    previousRoundFailedTests.remove(className, n -> {
```
pshevche, line 103: 💬 Actually, it would be enough to clear all failed tests for the entire class, since this lifecycle failure will trigger the retry of the entire class. But I think re-adding failed methods for retry is more semantically correct, and if those will not be retried for some other reason, we would still throw the same error.

plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGFuncTest.groovy
```diff
@@ -60,7 +60,7 @@ class TestNGFuncTest extends AbstractFrameworkFuncTest {
         with(result.output) {
             it.count('lifecycle FAILED') == 1
             it.count('lifecycle PASSED') == 1
-            !it.contains("org.gradle.test-retry was unable to retry")
+            !it.contains("The following test methods could not be retried")
```
pshevche, line 63: 💬 No similar test for TestNG, since in this case, it will still report descriptors for failed but not re-tried methods. They will be marked as skipped.

pshevche:
❗ This needs a test for exhausting all retries + what happens if setup fails after cleanup with test failures failed

PR #177, opened 2023-02-23
Handle class setup failures preventing previously failed methods from retrying

pshevche:
### Summary

If class setup fails, then test methods that failed in the previous execution won't be retried. In this case, we used to throw an error since we did not retry things that should be retried. This PR solves the issue by clearing the list of to-be-retried methods if class-level failure occurs. In this case, the entire class will be retried anyways, so it is safe to just clear that list. If previous run had other lifecycle errors, we track those to ensure we report successful execution of those.

### DIFF
--- a/plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
+++ b/plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
@@ -78,6 +78,24 @@ abstract class AbstractPluginFuncTest extends Specification {
                         throw new java.io.UncheckedIOException(e);
                     }
                 }
+
+                public static void flakyAssertPassFailPass(String id) {
+                    Path marker = Paths.get("build/marker.file." + id);
+                    try {
+                        if (Files.exists(marker)) {
+                            int counter = Integer.parseInt(new String(Files.readAllBytes(marker)));
+                            ++counter;
+                            Files.write(marker, Integer.toString(counter).getBytes());
+                            if (counter == 1) {
+                                throw new RuntimeException("fail me!");
+                            }
+                        } else {
+                            Files.write(marker, "0".getBytes());
+                        }
+                    } catch (java.io.IOException e) {
+                        throw new java.io.UncheckedIOException(e);
+                    }
+                }
             }
         """
     }
@@ -130,6 +148,10 @@ abstract class AbstractPluginFuncTest extends Specification {
 
     static String flakyAssert(String id = "id", int failures = 1) {
         return "acme.FlakyAssert.flakyAssert(\"${StringEscapeUtils.escapeJava(id)}\", $failures);"
+    }
+
+    static String flakyAssertPassFailPass(String id = "id") {
+        return "acme.FlakyAssert.flakyAssertPassFailPass(\"${StringEscapeUtils.escapeJava(id)}\");"
     }
 
     @SuppressWarnings("GroovyAssignabilityCheck")
--- a/plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
+++ b/plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
@@ -51,7 +51,7 @@ abstract class AbstractPluginFuncTest extends Specification {
     }
 
     String markerFileExistsCheck(String id = "id") {
-        "Files.exists(Paths.get(\"build/marker.file.${StringEscapeUtils.escapeJava(id)}\"))"
+        """Files.exists(Paths.get("build/marker.file.${StringEscapeUtils.escapeJava(id)}"))"""
     }
 
     String flakyAssertClass() {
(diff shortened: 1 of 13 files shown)


jean-andre-gauthier (approved):

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -78,6 +78,24 @@ abstract class AbstractPluginFuncTest extends Specification {
...
                     }
                 }
+
+                public static void flakyAssertPassFailPass(String id) {
+                    Path marker = Paths.get("build/marker.file." + id);
+                    try {
+                        if (Files.exists(marker)) {
+                            int counter = Integer.parseInt(new String(Files.readAllBytes(marker)));
+                            ++counter;
+                            Files.write(marker, Integer.toString(counter).getBytes());
```
leonard84, line 88:
```suggestion
                            marker.text = counter
```

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -78,6 +78,24 @@ abstract class AbstractPluginFuncTest extends Specification {
                         throw new java.io.UncheckedIOException(e);
                     }
                 }
+
+                public static void flakyAssertPassFailPass(String id) {
+                    Path marker = Paths.get("build/marker.file." + id);
+                    try {
+                        if (Files.exists(marker)) {
+                            int counter = Integer.parseInt(new String(Files.readAllBytes(marker)));
```
leonard84, line 86: You probably need the `groovy-nio` as test dependency for the nice `Path` extension methods.
```suggestion
                            int counter = Integer.parseInt(marker.text));
```

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -78,6 +78,24 @@ abstract class AbstractPluginFuncTest extends Specification {
...
+                            ++counter;
+                            Files.write(marker, Integer.toString(counter).getBytes());
+                            if (counter == 1) {
+                                throw new RuntimeException("fail me!");
+                            }
+                        } else {
+                            Files.write(marker, "0".getBytes());
```
leonard84, line 93:
```suggestion
                            marker.text = 0
```

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -78,6 +78,24 @@ abstract class AbstractPluginFuncTest extends Specification {
                         throw new java.io.UncheckedIOException(e);
                     }
                 }
+
+                public static void flakyAssertPassFailPass(String id) {
+                    Path marker = Paths.get("build/marker.file." + id);
```
leonard84, line 83:
```suggestion
                    Path marker = Paths.get("build/marker.file.$id");
```

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
```diff
@@ -132,6 +150,10 @@ abstract class AbstractPluginFuncTest extends Specification {
         return "acme.FlakyAssert.flakyAssert(\"${StringEscapeUtils.escapeJava(id)}\", $failures);"
     }
 
+    static String flakyAssertPassFailPass(String id = "id") {
+        return "acme.FlakyAssert.flakyAssertPassFailPass(\"${StringEscapeUtils.escapeJava(id)}\");"
```
leonard84, line 154: 🤔 you could use `"""` or as string delimiters to avoid having to use escapes for `\"`

leonard84 (approved):

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
pshevche: ❗ This is a Java class that we create in the build workspace to simplify creation of test classes, so this won't work.

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
pshevche: Same here, it is a plain Java class written alongside test classes.

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
pshevche: Ditto

plugin/src/test/groovy/org/gradle/testretry/AbstractPluginFuncTest.groovy
pshevche: Same

PR #193, opened 2023-05-24
Update plugin-publish plugin to 1.2.0

### LINKED ISSUE #179: Faulty `.module` file since version 1.3.0
Running `gradle build` on the following:

```
buildscript {
    repositories {
        repositories {
            maven {
                url 'https://plugins.gradle.org/m2'
                metadataSources {
                    gradleMetadata()
                    mavenPom()
                }
            }
        }

    }
    dependencies {
        classpath "org.gradle:test-retry-gradle-plugin:1.3.0"
    }
}

apply plugin: "org.gradle.test-retry"
```

Causes an error:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'gradle_retry_reproducer'.
> Could not resolve all files for configuration ':classpath'.
   > Could not find plugin-1.3.0.jar (org.gradle:test-retry-gradle-plugin:1.3.0).
     Searched in the following locations:
         https://plugins.gradle.org/m2/org/gradle/test-retry-gradle-plugin/1.3.0/plugin-1.3.0.jar
```

This appears to be because of the faulty jar file path `plugin-1.3.0.jar` included in the `.module` file: \
https://plugins.gradle.org/m2/org/gradle/test-retry-gradle-plugin/1.3.0/test-retry-gradle-plugin-1.3.0.module

The jar file path _should be_ `test-retry-gradle-plugin-1.3.0.jar`

# Notes:
- This error _does not_ occur with version 1.2.1 or below, because those versions do not have a `.module` file.
- This error _does_ occur with version 1.3.0 and above, including the latest version 1.5.2, because those versions all have a `.module` file with a faulty `.jar` path.
- This error does not occur when using `gradlePluginPortal()` as a repository, because it does not use `gradleMetadata()`
- Tested with Gradle versions {7.5.1, 8.1}

marcphilipp:
Resolves #179.
Closes #189.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -1,8 +1,7 @@
-import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import com.google.gson.Gson
 import org.gradle.testretry.build.GradleVersionData
 import org.gradle.testretry.build.GradleVersionsCommandLineArgumentProvider
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
-import com.google.gson.Gson
 import java.net.URL
 
 plugins {
@@ -44,9 +43,14 @@ tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions.jvmTarget = "1.8"
 }
 
-val plugin: Configuration by configurations.creating
+val plugin: Configuration by configurations.creating {
+    isCanBeConsumed = false
+    isCanBeResolved = true
+}
 
-configurations.getByName("compileOnly").extendsFrom(plugin)
+configurations.compileOnly {
+    extendsFrom(plugin)
+}
 
 dependencies {
     val asmVersion = "9.5"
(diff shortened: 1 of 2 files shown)


plugin/build.gradle.kts
```diff
@@ -124,11 +115,9 @@ license {
 
 publishing {
     publications {
-        register<MavenPublication>("plugin") {
+        create<MavenPublication>("pluginMaven") {
+            groupId = "org.gradle"
             artifactId = "test-retry-gradle-plugin"
-            artifact(shadowJar.get()) {
-                classifier = null
-            }
```
marcphilipp, line 131: @jbartok This should now be done automatically, right?

plugin/build.gradle.kts
```diff
@@ -124,11 +115,9 @@ license {
 
 publishing {
     publications {
-        register<MavenPublication>("plugin") {
+        create<MavenPublication>("pluginMaven") {
```
marcphilipp, line 118: @jbartok Is this the right way to customize the artifact ID and POM?

marcphilipp:
I published a snapshot to the dev plugin portal and it looks as it should: 🎉 
https://plugins.grdev.net/plugin/org.gradle.test-retry

plugin/build.gradle.kts
jbartok: Yes

plugin/build.gradle.kts
jbartok: Yes

jbartok (approved):

PR #200, opened 2023-07-06
Replace test class inheritence from a spec class with a common abstract spec class

pshevche:
With the current setup, the IDE won't be able to recognize that methods of `SpockFuncTest` can also be executed in the context of e.g. `Spock2FuncTest`. This happens because `SpockFuncTest` is not an abstract class and it can't be made one, because its own tests won't run. I introduced a common abstract base spec that is now shared by all children specs. This setup supports executing tests in the context of any child spec via IDE.

### DIFF
--- a/plugin/src/test/groovy/org/gradle/testretry/testframework/Spock1FuncTest.groovy
+++ b/plugin/src/test/groovy/org/gradle/testretry/testframework/Spock1FuncTest.groovy
@@ -15,42 +15,29 @@
  */
 package org.gradle.testretry.testframework
 
-abstract class SpockBaseJunit5FuncTest extends SpockBaseFuncTest {
-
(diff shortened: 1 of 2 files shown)


plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
marcphilipp, line 1: Should this class be called `Spock1FuncTest`?

marcphilipp (approved):

plugin/src/test/groovy/org/gradle/testretry/testframework/SpockFuncTest.groovy
pshevche: Makes sense, changed that

snoopcheri (approved):

PR #201, opened 2023-07-06
Do not activate plugin when --test-dry-run is set

jean-andre-gauthier:


plugin/src/test/groovy/org/gradle/testretry/TestDryRunFuncTest.groovy
```diff
@@ -0,0 +1,92 @@
+package org.gradle.testretry
+
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.util.GradleVersion
+
+class TestDryRunFuncTest extends AbstractGeneralPluginFuncTest {
+
+    private static final GradleVersion GRADLE_8_2_999 = GradleVersion.version("8.2.999")
+
+    def "emits skipped test method events if dryRun = true, retry is enabled, and TD/PTS are disabled"() {
```
pshevche, line 10: ❗ TD/PTS should not be mentioned in this context

pshevche (changes requested):

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -113,6 +114,17 @@ private static boolean supportsFinalizeValueOnRead(GradleVersion gradleVersion)
         return gradleVersion.compareTo(GRADLE_6_1) >= 0;
     }
 
+    private static boolean supportsDryRun(GradleVersion gradleVersion) {
+        return gradleVersion.compareTo(GRADLE_8_3) >= 0;
+    }
+
+    private static boolean callGetDryRun(GradleVersion gradleVersion, Test task) {
+        if (supportsDryRun(gradleVersion)) {
+            return invoke(declaredMethod(Test.class, "getDryRun"), task);
```
marcphilipp, line 123: ❌ Rather than using reflection, we should update the build to a Gradle nightly version and call it directly (behind the same if).

marcphilipp (changes requested):

marcphilipp (approved):

pshevche (approved):

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -54,14 +55,14 @@ public static void configureTestTask(Test test, ObjectFactory objectFactory, Pro
 
         test.getInputs().property("retry.failOnPassedAfterRetry", adapter.getFailOnPassedAfterRetryInput());
 
-        Provider<Boolean> isDeactivatedByTestDistributionPlugin =
+        Provider<Boolean> shouldReplaceTestExecutor =
```
snoopcheri, line 58: I would rename the method to `shouldReplaceTestExecutor()`.

marcphilipp (changes requested):
On second thought, not replacing it fro dry-run might not be necessary since the original `TestExecuter` it delegates to should already handle dry-run mode correctly. If that's the case, I think we should only keep the func test.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionSha256Sum=e111cb9948407e26351227dabce49822fb88c37ee72f1d1582a69c68af2e702f
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions-snapshots/gradle-8.3-20230709222231+0000-bin.zip
```
leonard84, line 4: Does this work without changing the `distributionSha256Sum` checksum?

plugin/src/test/groovy/org/gradle/testretry/TestDryRunFuncTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.util.GradleVersion
+
+class TestDryRunFuncTest extends AbstractGeneralPluginFuncTest {
+
+    private static final GradleVersion GRADLE_8_2_999 = GradleVersion.version("8.2.999")
+
+    def "emits skipped test method events if dryRun = true and retry plugin is enabled"() {
+        given:
+        def gradle83OrAbove = GradleVersion.version(gradleVersion) > GRADLE_8_2_999
```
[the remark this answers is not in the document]
jean-andre-gauthier: I had the issue that the nightly version was not considered `>= 8.3`, but this seems to be fine now, so I changed it to your proposition

jean-andre-gauthier:
> On second thought, not replacing it fro dry-run might not be necessary since the original `TestExecuter` it delegates to should already handle dry-run mode correctly. If that's the case, I think we should only keep the func test.

Makes sense, I've removed the dry-run logic, and only kept the test

gradle/wrapper/gradle-wrapper.properties
jean-andre-gauthier: Nope it didn't 😁  Fixed the checksum accordingly

plugin/src/main/java/org/gradle/testretry/internal/config/TestTaskConfigurer.java
```diff
@@ -54,14 +55,14 @@ public static void configureTestTask(Test test, ObjectFactory objectFactory, Pro
...
             shouldTestRetryPluginBeDeactivated(test, objectFactory, providerFactory, gradleVersion);
-        test.getInputs().property("isDeactivatedByTestDistributionPlugin", isDeactivatedByTestDistributionPlugin);
+        test.getInputs().property("isDeactivatedByTestDistributionPlugin", shouldReplaceTestExecutor);
```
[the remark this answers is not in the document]
jean-andre-gauthier: We'll keep the property as-is, as discussed on slack

jean-andre-gauthier:
rfm

PR #202, opened 2023-07-11
Retry Spock2 parameterized methods independently of the class

pshevche:
### Summary

Previously, if Spock tests were executed using Gradle's `JUnitPlatformTestFramework`, we would retry an entire class if one of its parameterized methods failed. This applied both to Spock 2 tests and Spock 1 tests run with JUnit Vintage  engine. However, parameterized tests cannot be retried independently only in the latter case.

This PR changes that and checks for `spock-core-2*` JAR on the classpath to differentiate between both cases. If the JAR is present, parameterized tests will be retried independently, otherwise, as an entire class.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
@@ -30,9 +30,15 @@ import static org.gradle.testretry.internal.executer.framework.TestFrameworkStra
 
 final class Junit5TestFrameworkStrategy extends BaseJunitTestFrameworkStrategy {
 
+    private final boolean isSpock2Used;
+
+    public Junit5TestFrameworkStrategy(boolean isSpock2Used) {
+        this.isSpock2Used = isSpock2Used;
+    }
+
     @Override
     public TestFramework createRetrying(TestFrameworkTemplate template, TestFramework testFramework, TestNames failedTests) {
-        DefaultTestFilter failedTestsFilter = testFilterFor(failedTests, false, template);
+        DefaultTestFilter failedTestsFilter = testFilterFor(failedTests, isSpock2Used, template);
         return testFrameworkProvider(template, testFramework).testFrameworkFor(failedTestsFilter);
     }
 
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
@@ -15,6 +15,7 @@
  */
 package org.gradle.testretry.internal.executer.framework;
 
+import org.gradle.api.internal.tasks.testing.JvmTestExecutionSpec;
 import org.gradle.api.internal.tasks.testing.TestFramework;
 import org.gradle.api.internal.tasks.testing.junit.JUnitTestFramework;
 import org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework;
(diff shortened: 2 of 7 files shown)


plugin/src/main/java/org/gradle/testretry/internal/executer/framework/Junit5TestFrameworkStrategy.java
```diff
@@ -30,9 +30,15 @@
...
+    private final boolean isSpock2Used;
+
+    public Junit5TestFrameworkStrategy(boolean isSpock2Used) {
+        this.isSpock2Used = isSpock2Used;
+    }
+
     @Override
     public TestFramework createRetrying(TestFrameworkTemplate template, TestFramework testFramework, TestNames failedTests) {
-        DefaultTestFilter failedTestsFilter = testFilterFor(failedTests, false, template);
+        DefaultTestFilter failedTestsFilter = testFilterFor(failedTests, isSpock2Used, template);
```
pshevche, line 41: 💬 This boolean will be propagated as`canRunParameterizedSpockMethods` to https://github.com/gradle/test-retry-gradle-plugin/blob/main/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/BaseJunitTestFrameworkStrategy.java#L91 that ultimately decides how to retry parameterized test failures: as a whole class or individually.

jean-andre-gauthier (approved):

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
```diff
@@ -25,25 +26,48 @@
...
+            return new Junit5TestFrameworkStrategy(isSpock2Used(spec));
         } else if (testFramework instanceof TestNGTestFramework) {
             return new TestNgTestFrameworkStrategy();
         } else {
             return null;
         }
     }
 
+    static boolean isSpock2Used(JvmTestExecutionSpec spec) {
+        Iterator<? extends File> classpathIterator = spec.getClasspath().iterator();
```
marcphilipp, line 55: ❓ Don't we also have to check `spec.getModulePath()` since Spock 2 should end up there if `java.modularity.inferModulePath` is set to `true` in the build script?

pshevche: It seems like setting it to `true` is not enough. Does it have to be a similar setup as described in https://docs.gradle.org/current/samples/sample_java_modules_multi_project_with_integration_tests.html? Where the test source set is a module itself?

pshevche: I added a check of module path in https://github.com/gradle/test-retry-gradle-plugin/pull/202/commits/d06b3752e9240e2f2497ca6fbd36e9b6ae50f674. But I have not managed to set up a test yet where that would happen. I can look into it once I am back.

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
```diff
@@ -25,25 +26,52 @@
...
+        Iterator<? extends File> classpathIterator = path.iterator();
+        boolean spock2CoreDependencyFound = false;
+        while (classpathIterator.hasNext() && !spock2CoreDependencyFound) {
+            File file = classpathIterator.next();
+            if (isSpock2CoreJar(file)) {
+                spock2CoreDependencyFound = true;
+            }
+        }
+
+        return spock2CoreDependencyFound;
```
leonard84, line 68:
```suggestion
        StreamSupport.stream(path.spliterator(), false)
            .anyMatch(this::isSpock2CoreJar);
```

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
```diff
@@ -25,25 +26,52 @@
...
+import java.io.File;
+import java.util.Iterator;
+import java.util.regex.Pattern;
 
 /**
  * Instances are scoped to a test task execution and are reused between rounds.
  */
 public interface TestFrameworkStrategy {
 
+    Pattern SPOCK2_CORE_JAR_NAME_PATTERN = Pattern.compile(".*/spock-core-2.*\\.jar");
```
leonard84, line 38:
```suggestion
    Pattern SPOCK2_CORE_JAR_NAME_PATTERN = Pattern.compile(".*/spock-core-2[^/]*\\.jar");
```

leonard84 (approved):

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
pshevche: Done in https://github.com/gradle/test-retry-gradle-plugin/pull/202/commits/e03e80d0e4981334234d61187dc28f19da1ededb

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/TestFrameworkStrategy.java
pshevche: TIL. Done in https://github.com/gradle/test-retry-gradle-plugin/pull/202/commits/e03e80d0e4981334234d61187dc28f19da1ededb

PR #231, opened 2023-11-01
Fix class-level retry with Gradle 5.0 and Suite engine or @Nested test classes

### LINKED ISSUE #230: Test retry on class level fails with suite engine and Gradle 5.0
When using the Test Retry Gradle plugin with Gradle 5.0 and using the JUnit Suite engine, retrying tests on class level fails.

```
TestSuite
  Test1
    testOk()
    testFlaky() // fails on 1st execution and passes on 2nd
  Test2
    testOk()
    testFlaky() // fails on 1st execution and passes on 2nd

```

The class `Test1` is configured to be retried on class-level.

When running this with Gradle 5.0, the following error is shown:

```
The following test methods could not be retried, which is unexpected.
```

No test methods are listed though.

The reason for this problem is the fact that Gradle 5.0 does not report test events for the intermediate classes `Test1` and `Test2`. Only events for `TestSuite` and the test methods are reported.

snoopcheri:
This PR adds special handling when Gradle 5.0 is used.

At the end of each round, entries for test classes w/o test methods from `previousRoundFailedTests`, when using Gradle 5.0. For Gradle 5.1 and above this would automatically happen, as Gradle also reports events for intermediate classes, when JUnit's suite engine is used or when test classes are `@nested`.

To be on the safe side, we're only removing entries for test classes, if we observed this test class during the current round. It's not a perfect solution but allows using the class-retry feature of the Test Retry Gradle plugin together with test suites and nested class for Gradle 5.0 as well.

Fixes #230

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -25,10 +25,13 @@ import org.gradle.testretry.internal.executer.framework.TestFrameworkStrategy;
 import org.gradle.testretry.internal.filter.ClassRetryMatcher;
 import org.gradle.testretry.internal.filter.RetryFilter;
 import org.gradle.testretry.internal.testsreader.TestsReader;
+import org.gradle.util.GradleVersion;
 
 import java.lang.reflect.Method;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import static org.gradle.api.tasks.testing.TestResult.ResultType.SKIPPED;
 
@@ -47,6 +50,7 @@ final class RetryTestResultProcessor implements TestResultProcessor {
 
     private final Map<Object, TestDescriptorInternal> activeDescriptorsById = new HashMap<>();
 
+    private final Set<String> testClassesSeenInCurrentRound = new HashSet<>();
     private TestNames currentRoundFailedTests = new TestNames();
     private TestNames previousRoundFailedTests = new TestNames();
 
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -207,7 +220,53 @@ private boolean currentRoundFailedTestsExceedsMaxFailures() {
...
+     * <p/>
+     * This solution is not perfect but still allows users to use classRetry with Gradle 5
+     * together with the suite engine and/or nested test classes.
+     *
+     * @return cleaned up failed test names of previous round
+     */
+    private TestNames cleanedUpFailedTestsOfPreviousRound() {
+        boolean isGradle50 = GradleVersion.current().getBaseVersion().equals(GradleVersion.version("5.0"));
+
+        if (isGradle50 && !testClassesSeenInCurrentRound.isEmpty() || previousRoundFailedTests.hasClassesWithoutTestNames()) {
```
pshevche, line 251: ❓ Shouldn't it be `isGradle50 && !testClassesSeenInCurrentRound.isEmpty() && previousRoundFailedTests.hasClassesWithoutTestNames()`? If I understand it correctly, we will enter this block in the following cases when it is not necessary:

* `isGradle50 = false && previousRoundFailedTests.hasClassesWithoutTestNames()` -> we will simply recreate the `previousRoundFailedTests`
* `isGradle50 = true && !testClassesSeenInCurrentRound.isEmpty() && hasClassesWithoutTestNames = false`. In this case, we will never enter the L257 branch and also simply recreate the `previousRoundFailedTests`.

pshevche (changes requested):

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
snoopcheri: You're right.

I think I was playing with the idea to convert the condition from `!a && !b` to `!(a || b)` and didn't properly revert it.

FIXED (aad6d860)

pshevche (approved):

PR #273, opened 2024-04-09
Account for different formats of reported display name

pshevche:
Gradle 8.8+ will propagate the reported display name from JUnit. This change updates impacted assertions to work both with new and old output formats. See https://e.grdev.net/s/stn5aqybyzdt2/tests/task/:plugin:testGradleNightlies/details/org.gradle.testretry.FilterFuncTest/can%20filter%20what%20is%20retried%20(gradle%20version%208.8-20240406001318%2B0000)?top-execution=1


plugin/src/test/groovy/org/gradle/testretry/FilterFuncTest.groovy
```diff
@@ -56,14 +56,14 @@ class FilterFuncTest extends AbstractGeneralPluginFuncTest {
         then:
         noRetry.each { testName ->
             with(result.output) {
-                it.count("acme.${testName} > flakyTest FAILED") == 1
```
pshevche, line 59: 💬 Gradle 8.7 and older reports `acme.ExcludedTest`. Gradle 8.8+ reports `ExcludedTest`

skuzzle (approved):

PR #296, opened 2024-07-11
Support JUnit TestNG engine

pshevche:
## Summary

SAP reports a number of issues when TestNG run with [JUnit TestNG Engine](https://github.com/junit-team/testng-engine) are retried:
1. Methods with dependencies are not retried.
2. Lifecycle failure retries are reported as unsuccessful.
3. All these failures result in a Build Scan not being published.

This PR addresses the first two and I manually verified that the fix also resolves Build Scan publication errors. The plan forward is to release a Develocity plugin RC with this fix and validate with the customer if all of the issues are resolved.

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -33,7 +33,9 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.toMap;
 import static org.gradle.api.tasks.testing.TestResult.ResultType.SKIPPED;
 
 final class RetryTestResultProcessor implements TestResultProcessor {
@@ -91,6 +93,10 @@ final class RetryTestResultProcessor implements TestResultProcessor {
     @Override
     public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
         if (testId.equals(rootTestDescriptorId)) {
+            // nothing failed in the current round, but we have some un-retried tests
+            if (currentRoundFailedTests.isEmpty() && !previousRoundFailedTests.isEmpty()) {
+                ignoreExpectedUnretriedTests();
+            }
             if (!lastRun()) {
                 return;
             }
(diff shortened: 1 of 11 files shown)


plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -127,13 +133,24 @@ public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
...
+    private void ignoreExpectedUnretriedTests() {
+        // check with the framework implementation if it is expected
+        Map<String, Set<String>> expectedUnretriedTests = previousRoundFailedTests.stream()
+            .collect(toMap(
+                Map.Entry::getKey,
+                entry -> entry.getValue().stream()
+                    .filter(test -> testFrameworkStrategy.isExpectedUnretriedTest(entry.getKey(), test))
+                    .collect(Collectors.toSet())
+            ));
+        expectedUnretriedTests.forEach((className, tests) -> previousRoundFailedTests.remove(className, tests::contains));
```
pshevche, line 151: 💬 Ideally, we would also report failed descriptors as successful, but at this point, we have no clue where in the hierarchy they belong or how to construct a valid descriptor.

With all of the issues with the TestNG reporting, it seems like putting lipstick on a pig. So, I suggest not to bother and just make sure that retries work correctly.

plugin/src/test/groovy/org/gradle/testretry/testframework/BaseTestNGFuncTest.groovy
```diff
@@ -205,7 +244,7 @@ class TestNGFuncTest extends AbstractFrameworkFuncTest {
 
             public class OrderedTests {
                 @Test(dependsOnMethods = {"childTest"})
-                public void grandchildTest() {}
+                public void grandChildTest() {}
```
jean-andre-gauthier, line 247: Afaik grandchild is a single word, so the case could be made to leave the casing as-is

jean-andre-gauthier (approved):

plugin/src/test/groovy/org/gradle/testretry/testframework/BaseTestNGFuncTest.groovy
pshevche: Agree, it is hard to make assertions, though, because `childTest` matches both the `childTest` and the `grandchildTest`.

PR #300, opened 2024-07-16
Fix assertions on parameterized test names for older Gradle versions

pshevche:
Fixing https://e.grdev.net/s/doszanvifrfno/tests/overview?outcome=FAILED

### DIFF
--- a/plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGViaJUnitEngineFuncTest.groovy
+++ b/plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGViaJUnitEngineFuncTest.groovy
@@ -32,7 +32,10 @@ class TestNGViaJUnitEngineFuncTest extends BaseTestNGFuncTest {
     private static final EnumSet<TestNGLifecycleType> CLASS_LIFECYCLE_METHODS = EnumSet.of(BEFORE_CLASS, BEFORE_METHOD, AFTER_METHOD)
 
     private static final GradleVersion GRADLE_5_0 = GradleVersion.version("5.0")
-    private static final GradleVersion GRADLE_5_4_1 = GradleVersion.version("5.4.1")
+    private static final GradleVersion GRADLE_6_1 = GradleVersion.version("6.1")
+    private static final GradleVersion GRADLE_7_0 = GradleVersion.version("7.0")
+    private static final GradleVersion GRADLE_7_6_4 = GradleVersion.version("7.6.4")
+    private static final GradleVersion GRADLE_8_1 = GradleVersion.version("8.1")
 
     def setup() {
         buildFile << """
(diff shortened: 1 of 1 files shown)


plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGViaJUnitEngineFuncTest.groovy
```diff
@@ -55,9 +58,12 @@ class TestNGViaJUnitEngineFuncTest extends BaseTestNGFuncTest {
 
     @Override
     String reportedParameterizedMethodName(String gradleVersion, String methodName, String paramType, int invocationNumber, @Nullable String paramValueRepresentation) {
-        GradleVersion.version(gradleVersion) > GRADLE_5_4_1
-            ? "${methodName}(${paramType}) > [${invocationNumber}] ${paramValueRepresentation ?: ''}"
-            : "${methodName}(${paramType})[${invocationNumber}]"
+        switch (GradleVersion.version(gradleVersion)) {
+            case { it < GRADLE_6_1 }: return "${methodName}(${paramType})[${invocationNumber}]"
+            case { it < GRADLE_7_0 }: return "[${invocationNumber}] ${paramValueRepresentation ?: ''}"
+            case { it < GRADLE_8_1 && it != GRADLE_7_6_4 }: return "${methodName}(${paramType})[${invocationNumber}]"
```
snoopcheri, line 64: I assume that a potential Gradle 7.6.5 release would also match the pattern in line 64, but I have no idea how likely such a version would be. So it's probably not worth considering it at the moment.

snoopcheri (approved):

plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGViaJUnitEngineFuncTest.groovy
jean-andre-gauthier, line 64: Wouldn't it be more resilient to use `it < GRADLE_8_1 && !(it >= GRADLE_7_6_4 && it < GRADLE_8_0)` as a condition?

jean-andre-gauthier (approved):

plugin/src/test/groovy/org/gradle/testretry/testframework/TestNGViaJUnitEngineFuncTest.groovy
pshevche: Yeah, I think we'll notice it during our CI checks and can easily fix it then.

pshevche: I doubt there will be another 7.6.x release, so let's try to keep it simple for now.

PR #302, opened 2024-07-23
Ignore tests not compatible with Gradle 5.x

pshevche:


plugin/src/test/groovy/org/gradle/testretry/ParenthesesFuncTest.groovy
```diff
@@ -40,8 +49,7 @@ class ParenthesesFuncTest extends AbstractPluginFuncTest {
 
         where:
         [gradleVersion, scenarios] << [
-            // Kotlin plugin compatible from 6.8 onwards
-            GRADLE_VERSIONS_UNDER_TEST.findAll { GradleVersion.version(it) >= GradleVersion.version("6.8") },
```
pshevche, line 44: 💬 After splitting the nightly task into multiple based on the Gradle version, the `testGradle5Releases` task will result in no data for `gradleVersion`, causing the test to fail.

jean-andre-gauthier (approved):

Issue #130, opened 2022-04-12
Shouldn't it rather be failOnPassedOrSkippedAfterRetry?


ivans083:
If a test is set to be skipped on it's body if certain condition meets, a retried test that is later skipped will fail the build even if failOnPassedAfterRetry = false.
We could either fix and call it failOnPassedOrSkippedAfterRetry
Or add failOnSkippedAfterRetry option in configuration

buitcj:
I also would love this

adboyarshinov:
In my case, tests cases, that have known bugs are marked with special method `knownBUG("BUG-123")`, that set test to skipped and link issue `BUG-123` to report.
So in test report all skipped tests = known bugs.
Problem occurs then fail accidentally happens before `knownBug("BUG-123")`, and in retry successfully reaches this method.

pshevche:
@adboyarshinov , thank you for clarifying the use case. We've discussed it internally, and your proposal makes sense. Let us continue with the PR. Ping me on the PR if you need any help from my side.

PR #313, opened 2024-09-09
Add failOnSkippedAfterRetry option

adboyarshinov:
If a test is set to be skipped on it's body if certain condition meets, a retried test that is later skipped will fail the build.
Adding this option allows you to choose the behavior you want.

There is issue for this problem:
[#130](https://github.com/gradle/test-retry-gradle-plugin/issues/130)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -47,6 +47,7 @@ test {
         maxRetries = 2
         maxFailures = 20
         failOnPassedAfterRetry = true
+        failOnSkippedAfterRetry = true
     }
 }
 ----
@@ -60,6 +61,7 @@ test {
         maxRetries.set(2)
         maxFailures.set(20)
         failOnPassedAfterRetry.set(true)
+        failOnSkippedAfterRetry.set(true)
     }
 }
 ----
@@ -79,6 +81,7 @@ test {
             maxFailures = 20
         }
         failOnPassedAfterRetry = true
+        failOnSkippedAfterRetry = true
     }
 }
 ----
@@ -119,6 +122,18 @@ public interface TestRetryTaskExtension {
      * @return whether tests that initially fails and then pass on retry should fail the task
      */
     Property<Boolean> getFailOnPassedAfterRetry();
+
+    /**
+     * Whether tests that initially fail and then are skipped on retry should fail the task.
+     * <p>
+     * This setting defaults to {@code true} (for backward compatibility),
+     * which results in the task failing if any of tests skip on retry.
+     * <p>
+     * This setting has no effect if {@link Test#getIgnoreFailures()} is set to true.
+     *
+     * @return whether tests that initially fails and then are skipped on retry should fail the task
+     */
+    Property<Boolean> getFailOnSkippedAfterRetry();
 
     /**
      * The maximum number of times to retry an individual test.
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -47,6 +47,7 @@ final class RetryTestResultProcessor implements TestResultProcessor {
     private final TestResultProcessor delegate;
 
     private final int maxFailures;
+    private final boolean failOnSkippedAfterRetry;
     private boolean lastRetry;
     private boolean hasRetryFilteredFailures;
     private Method failureMethod;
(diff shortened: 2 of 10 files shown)


pshevche:
@adboyarshinov, thank you for the PR. Could you please chime in on [this question](https://github.com/gradle/test-retry-gradle-plugin/issues/130#issuecomment-2032129688) from the issue? The original reporters did not follow up on it, and I am trying to understand the use case better before we commit to introducing this option. Do you have an example of such a volatile precondition that would cause tests to run, fail, and then be skipped?

README.adoc
```diff
@@ -120,6 +123,18 @@ public interface TestRetryTaskExtension {
      */
     Property<Boolean> getFailOnPassedAfterRetry();
 
+    /**
+     * Whether tests that initially fail and then skip on retry should fail the task.
```
pshevche, line 127:
```suggestion
     * Whether tests that initially fail and then are skipped on retry should fail the task.
```

Also in the other places.

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -108,7 +111,12 @@ public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
 
                 boolean failedInPreviousRound = previousRoundFailedTests.remove(className, name);
                 if (failedInPreviousRound && testCompleteEvent.getResultType() == SKIPPED) {
```
pshevche, line 113: I suggest collapsing these two conditions into a single boolean property to better convey the intent. Maybe something like this:

```
boolean shouldRetrySkippedTestThatPreviouslyFailed = failedInPreviousRound && testCompleteEvent.getResultType() == SKIPPED && failOnSkippedAfterRetry;
if (shouldRetrySkippedTestThatPreviouslyFailed) {
   addRetry(descriptor);
}
```

plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -108,7 +111,12 @@ public void completed(Object testId, TestCompleteEvent testCompleteEvent) {
...
-                    addRetry(descriptor);
+                    if (failOnSkippedAfterRetry) {
+                        addRetry(descriptor);
+                    } else {
+                        emitFakeSkippedEvent(descriptor, testCompleteEvent, name);
```
pshevche, line 117: I don't think you need this `else`, because Gradle should report the skipped execution. I think this will cause the method to be reported twice (for example, in the console log).

pshevche (changes requested):
All in all, looks good, well done! I left a couple of comments and suggest the following:
1. Let us discuss how exactly this helps. I am still struggling to wrap up my head around the use case.
2. Please add a test for this. You can use [this test](https://github.com/gradle/test-retry-gradle-plugin/blob/05138d3d54d6347c254beb40ac333637d96c13ab/plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit4FuncTest.groovy#L346) as an example, but I'd recommend adding one to `JUnit5FuncTest`.

plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit4FuncTest.groovy
```diff
@@ -375,6 +375,39 @@ class JUnit4FuncTest extends AbstractFrameworkFuncTest {
         gradleVersion << GRADLE_VERSIONS_UNDER_TEST
     }
 
+    def "test that is skipped after failure with option 'failOnSkippedAfterRetry = false' is passes (gradle version #gradleVersion)"() {
```
pshevche, line 378: Please also add a test checking the default behavior that retries will fail.

pshevche (commented):
Let's add a test for the default behavior and we are good to go!

plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit4FuncTest.groovy
adboyarshinov: it is presented in [test above](https://github.com/gradle/test-retry-gradle-plugin/blob/99cb2792f6c0020e825a71f37a879f9526e81169/plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit4FuncTest.groovy#L346)

pshevche: Ah, right, missed that. Looks good then.

pshevche (approved):

pshevche:
Thank you for your contribution, @adboyarshinov . We'll try to release it in 1.6.0 as soon as possible, probably this week.

PR #348, opened 2024-10-30
Fixes broken :testGradle...Releases tasks

jean-andre-gauthier:


plugin/build.gradle.kts
```diff
@@ -156,7 +156,7 @@ tasks.withType<Test>().configureEach {
     useJUnitPlatform()
 }
 
-tasks.test {
+tasks.withType<Test> {
```
jean-andre-gauthier, line 159: This ensures that the system properties are available for all test tasks, not just the default one

Issue #209, opened 2023-09-03
Disable parallelism on retry


nedtwigg:
I have a big bank of tests where:

- 100% pass if we run the testsuite single-threaded
- 100% *sometimes* pass if we run the testsuite in parallel
- 5% of tests on average will fail if we run the testsuite in parallel, due to heavy load

This is good - we can speed up our test suite a ton by going parallel, and then rerun the handful of tests that failed due to load.

But ideally I wish the rerun could do something like:

```gradle
test {
  useJUnitPlatform {}
  systemProperty 'junit.vintage.execution.parallel.enabled', 'true'
  retry {
    maxRetries = 1
    onRetry {
      systemProperty 'junit.vintage.execution.parallel.enabled', 'false'
    }
  }
}
```

Issue #206, opened 2023-08-10
Run command before retry or before last retry


oeresundsgruppen:
Would it be an idea to allow to run a command before a retry or before last retry, e.g. restart a service, reload db etc.

CommandBeforeRetry=sqlcmd delete from tableX
CommandBeforeLastRetry=gradlew restartX && wait 60

Maybe I could add this and make PR for your review ?

oeresundsgruppen:
This is beneficial when tests are not mackovian, e.g. they built up state...

PR #349, opened 2024-11-08
Add onRetry for setting JVM options for fork differently

### LINKED ISSUE #205: Is there a way to output something when the retry succeeds?
I have a huge test suite and occasionally on CI builds my test runners go out of memory and fail. This plugin is a great stopgap while I try to diagnose the memory issues.

However, it's super confusing for developers who are casually working in the codebase - the build output makes it look like the test failed. Or worse, there might be 10 retryable failures and one "real" failure and it's hard to pinpoint the actual problem.

Here's some sample output from a recent build:

```
> Task :test
DeleteActionJspTest > test_content() FAILED
    java.lang.RuntimeException: java.io.IOException: Server returned HTTP response code: 500 for URL: http://localhost:38675/WEB-INF/views/mobile/facets/actions/DeleteAction.jsp
        at jsp.test.framework.core.JspServer.render(JspServer.java:71)
        at jsp.test.framework.core.JspTestCase.renderJspToElement(JspTestCase.java:111)
        at webapp.testframework.CartJspTestCase.lambda$runTests$6(CartJspTestCase.java:174)
        at com.google.common.collect.RegularImmutableMap.forEach(RegularImmutableMap.java:185)
        at webapp.testframework.CartJspTestCase.runTests(CartJspTestCase.java:172)
        at webapp.testframework.CartJspTestCase.runSharedTestsWithDevice(CartJspTestCase.java:163)
        at webinf.views.common.facets.actions.DeleteActionJspTest.test_content(DeleteActionJspTest.java:27)
        Caused by:
        java.io.IOException: Server returned HTTP response code: 500 for URL: http://localhost:38675/WEB-INF/views/mobile/facets/actions/DeleteAction.jsp
            at sun.net.[www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1902)](http://www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1902))
            at sun.net.[www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1500)](http://www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1500))
            at jsp.test.framework.core.JspServer.render(JspServer.java:69)
            ... 6 more
6934 tests completed, 1 failed, 3 skipped
There were failing tests. See the report at: file:///local/p4clients/pkgbuild-vU1LG/workspace/src/CartExperienceWebapp/build/brazil-unit-tests/index.html

> Task :preCoverageReport
```

it would be great if I could insert a line right after the failure saying something like "DeleteActionJspTest passed on retry #1" or similar. Is that already possible? If not, how hard is it to code up?

(would also be nice to be able to print a "10 tests passed on retry" or similar in the summary part)

I'm open to contributing the code if you are aligned with this being a good idea. Let me know, thanks!

U1F984:
Fixes #205, #206, #209, #271.

Going by my last submitted PR I don't expect this one to be accepted either, so I've not bothered to add any tests/docs.


pshevche:
@U1F984 , thank you for looking into it. I am not convinced that we'll go down this route. It does not cover some of the use cases from #271, like "Print a message on successful retry". Furthermore, we need to think more about the semantics of this callback: should it be invoked once, meaning that the test retry plugin kicked in, or on every retry? I am sorry it takes us so long to get to this topic. We still plan to investigate it, but we'll need to see first how it ranks against other priorities.

PR #365, opened 2025-01-17
Add support for dynamic tests

jean-andre-gauthier:
Updates `PARAMETERIZED_SUFFIX_PATTERN` in `BaseJunitTestFrameworkStrategy` to support dynamic tests, by allowing an arbitrary number of trailing `[...]`

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/BaseJunitTestFrameworkStrategy.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/framework/BaseJunitTestFrameworkStrategy.java
@@ -159,10 +159,11 @@ abstract class BaseJunitTestFrameworkStrategy implements TestFrameworkStrategy {
 
     private void addPotentiallyParameterizedSuffixed(TestFilterBuilder filters, String className, String name) {
         // It's a common pattern to add all the parameters on the end of a literal method name with []
-        // The regex takes care of removing trailing (...) or (...)[...], for e.g. the following cases
+        // The regex takes care of removing trailing (...) or (...)[...], or (...)[...][...] for e.g. the following cases
         // * `test that contains (parentheses)()`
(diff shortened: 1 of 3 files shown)


plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit5FuncTest.groovy
```diff
@@ -718,6 +719,63 @@ class JUnit5FuncTest extends AbstractFrameworkFuncTest {
...
+            }
+        """
+
+        when:
+        def result = gradleRunner(gradleVersion).build()
+
+        then:
+        def gv = GradleVersion.version(gradleVersion)
+        def testname
+        if (gv >= GradleVersion.version("8.8")) {
```
pshevche, line 756: ❔ Can we extract it into a method like `dynamicTestName(version, name)` that would just return the string we should assert on?

plugin/src/main/java/org/gradle/testretry/internal/executer/framework/BaseJunitTestFrameworkStrategy.java
```diff
@@ -35,7 +35,7 @@
 abstract class BaseJunitTestFrameworkStrategy implements TestFrameworkStrategy {
 
     public static final Logger LOGGER = LoggerFactory.getLogger(JunitTestFrameworkStrategy.class);
-    private static final Pattern PARAMETERIZED_SUFFIX_PATTERN = Pattern.compile("(?:\\([^)]*?\\))?(?:\\[[^]]*?])?$");
+    private static final Pattern PARAMETERIZED_SUFFIX_PATTERN = Pattern.compile("(?:\\([^)]*?\\))?(?:\\[[^]]*?])*$");
```
pshevche, line 38: 💭 Sometimes I have a feeling that the entire plugin relies on this one regex :D

pshevche (approved):

PR #375, opened 2025-02-05
Configure Git credential helper for tagging the release

snoopcheri:
For quite some time, tagging the repository with the new release version fails when releasing the Test Retry Gradle plugin.

The reason for this is that the [nebula-release-plugin](https://github.com/nebula-plugins/nebula-release-plugin) no longer uses GrGit or JGit, and therefore nobody reads the environment variables `GIT_USERNAME` or `GRGIT_USERNAME`.

This PR configures a temporary credential helper for Git before actually releasing the plugin. This should work, but we will only know when we try it.

I've also upgraded to the latest Kotlin DSL for TeamCity, because I thought I would need it. That's probably not the case, but it still makes sense to do this, so I kept it.

Verify All build: [here](https://builds.gradle.org/buildConfiguration/OpenSourceProjects_GradleTestRetryPlugin_null_VerifyAll/94448310)

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -1,12 +1,12 @@
-import jetbrains.buildServer.configs.kotlin.v2019_2.*
-import jetbrains.buildServer.configs.kotlin.v2019_2.BuildTypeSettings.Type.COMPOSITE
-import jetbrains.buildServer.configs.kotlin.v2019_2.FailureAction
-import jetbrains.buildServer.configs.kotlin.v2019_2.ParameterDisplay.NORMAL
-import jetbrains.buildServer.configs.kotlin.v2019_2.ParameterDisplay.PROMPT
-import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.gradle
-import jetbrains.buildServer.configs.kotlin.v2019_2.project
-import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.schedule
-import jetbrains.buildServer.configs.kotlin.v2019_2.version
+import jetbrains.buildServer.configs.kotlin.*
+import jetbrains.buildServer.configs.kotlin.BuildTypeSettings.Type.COMPOSITE
+import jetbrains.buildServer.configs.kotlin.FailureAction
+import jetbrains.buildServer.configs.kotlin.ParameterDisplay.NORMAL
+import jetbrains.buildServer.configs.kotlin.ParameterDisplay.PROMPT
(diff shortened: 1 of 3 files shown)


.teamcity/settings.kts
```diff
@@ -1,12 +1,13 @@
-import jetbrains.buildServer.configs.kotlin.v2019_2.*
-import jetbrains.buildServer.configs.kotlin.v2019_2.BuildTypeSettings.Type.COMPOSITE
-import jetbrains.buildServer.configs.kotlin.v2019_2.FailureAction
-import jetbrains.buildServer.configs.kotlin.v2019_2.ParameterDisplay.NORMAL
-import jetbrains.buildServer.configs.kotlin.v2019_2.ParameterDisplay.PROMPT
-import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.gradle
-import jetbrains.buildServer.configs.kotlin.v2019_2.project
-import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.schedule
-import jetbrains.buildServer.configs.kotlin.v2019_2.version
+import jetbrains.buildServer.configs.kotlin.*
```
snoopcheri, line 1: The new Kotlin DSL doesn't use the version name in the package.

.teamcity/settings.kts
```diff
@@ -226,12 +227,22 @@ project {
                 password("env.PGP_SIGNING_KEY_PASSPHRASE", "%pgpSigningPassphrase%")
             }
             steps {
+                script {
+                    scriptContent = """
+                        git config credential.helper '!f() { sleep 1; echo "username=${'$'}{GIT_USERNAME}"; echo "password=${'$'}{GIT_ACCESS_TOKEN}"; }; f'
```
snoopcheri, line 232: This is a custom credential helper. The `!` is a special syntax from Git and means that it should be executed in a Shell.

pshevche: What is the purpose of `sleep 1`? :)

pshevche (approved):

.teamcity/settings.kts
snoopcheri: I saw this solution [here](https://stackoverflow.com/questions/8536732/can-i-hold-git-credentials-in-environment-variables), where they claimed that on some systems this resolves a race condition. Nobody seems to know, though, who is actually racing. :)

I think it has to do with the fact that stdin is typically drained before a password prompt. That's why you cannot do things like `echo my_password | execute_something_which_is_requesting_a_password`. But I'm just guessing here.

PR #405, opened 2025-05-26
Use JDK17 to run tests against Gradle 9.0 nightlies

pshevche:
Gradle 9.0 requires at least JDK 17 to execute the test builds. This change uses the existing property we have in place to overwrite the Java version used for compilation and tests and sets it to 17 when tests use Gradle nightlies.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -36,6 +36,7 @@ version = "2024.12"
 project {
     params {
         java8Home(Os.linux)
+        java17Home(Os.linux)
         text("systemProp.org.gradle.internal.publish.checksums.insecure", "true")
     }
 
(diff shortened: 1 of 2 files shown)


.teamcity/settings.kts
```diff
@@ -36,6 +36,7 @@ version = "2024.12"
 project {
     params {
         java8Home(Os.linux)
+        java17Home(Os.linux)
```
wolfs, line 39: 🤔 I don't understand why you need to do this. The OS configuration should take care of it already: https://github.com/gradle/dv/blob/4dcb421c32bc668e2891d429ecf2501ae918de17/.teamcity/src/main/kotlin/com/gradle/teamcity/dsl/DslExtensions.kt#L967-L984

snoopcheri (approved):

PR #450, opened 2025-09-17
Fix bug in TestsReader

kilink:
Use `Optional.ofNullable` to avoid throwing an NPE when a class file can't be read.

`visitClassFile` is annotated with `@Nullable`, and reading the code I can see that it returns null if it's unable to parse the class file. We ran into this issue when running tests on JDK 25. I guess it's indicative that Java 25 class files can't be parsed, but the intent of the code is to log a warning and proceed as far as I can tell.


snoopcheri (approved):

kilink:
@snoopcheri can you take a look now? I've fixed the issue blocking the mergeability. Thanks!

PR #518, opened 2026-05-07
Restore compatibility with Gradle 9.3+

oehme:
See individual commits for details

### DIFF
--- a/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
+++ b/plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
@@ -176,7 +176,7 @@ final class RetryTestResultProcessor implements TestResultProcessor, Closeable {
     private void registerSeenTestClass(TestDescriptorInternal descriptor) {
         String maybeTestClassName = descriptor.getClassName();
 
-        if (maybeTestClassName != null) {
+        if (maybeTestClassName != null && !maybeTestClassName.isEmpty()) {
             testClassesSeenInCurrentRound.add(maybeTestClassName);
         }
     }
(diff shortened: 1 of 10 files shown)


plugin/src/main/java/org/gradle/testretry/internal/executer/RetryTestResultProcessor.java
```diff
@@ -253,12 +267,14 @@ private void failure(Object testId) {
         final TestDescriptorInternal descriptor = activeDescriptorsById.get(testId);
         if (descriptor != null) {
             String className = descriptor.getClassName();
-            if (className != null) {
+            if (className != null && !className.isEmpty()) {
```
pshevche, line 270: Can you please incorporate [this test](https://github.com/gradle/test-retry-gradle-plugin/pull/501/changes#diff-eac149948cb97de3067bd5f738553488ffc0d19ee03858bda972d33e2d0d4b18) from [this PR](https://github.com/gradle/test-retry-gradle-plugin/pull/501)? It looks like your fix should cover that issue, too.

pshevche (commented):
The changes look good, thank you for a quick turnaround. I left a minor comment regarding extending the test coverage and that would be it.

plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit5FuncTest.groovy
```diff
@@ -520,6 +520,49 @@ class JUnit5FuncTest extends AbstractFrameworkFuncTest {
...
+            }
+        """
+
+        when:
+        def result = gradleRunner(gradleVersion).buildAndFail()
+
+        then:
+        with(result.output) {
+            it.contains("> There were failing tests.")
+            it.count("${beforeClassErrorTestMethodName(gradleVersion)} FAILED") == 3
```
pshevche: 0: can we somehow adjust the test to succeed on retries? It is not a dealbreaker if not. It looks pretty tricky.

pshevche (approved):

plugin/src/test/groovy/org/gradle/testretry/testframework/JUnit5FuncTest.groovy
oehme: Mhhh that's a good point, almost nothing can fail discovery unless it's a real failure (e.g. classloading). So I guess retrying is really not what we want here. How about we keep this test as the "current state" and make "don't retry discovery issues" a separate issue?

pshevche: Yes, that's absolutely fine. Would you be able to take a look at it to wrap up this piece of work?

oehme: Yes, I will. Also gotta look at the compile errors with the Gradle nightly

PR #523, opened 2026-05-22
fix: declare the CC compatibility

pshevche:
The publication fails with 

```
Execution failed for task ':plugin:publishPlugins'.
06:56:53   > Failed to post to server.
06:56:53     Server responded with:
06:56:53     The following feature compatibility declarations were present in the previous version but are missing in the new version: 'configuration-cache'. Feature compatibility declarations cannot be removed once they have been added.
```

This is because the Gradle-owned plugins were retrospectively marked as CC-compatible without updating the publication code.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -1,4 +1,5 @@
 import com.google.gson.Gson
+import org.gradle.plugin.compatibility.compatibility
 import org.gradle.testretry.build.GradleVersionData
 import org.gradle.testretry.build.GradleVersionsCommandLineArgumentProvider
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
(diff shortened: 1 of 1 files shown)


plugin/build.gradle.kts
```diff
@@ -99,12 +100,17 @@ gradlePlugin {
     website.set("https://github.com/gradle/test-retry-gradle-plugin")
     vcsUrl.set("https://github.com/gradle/test-retry-gradle-plugin.git")
     plugins {
-        register("testRetry") {
+        create("testRetry") {
```
snoopcheri: ❔Why the change from `register` to `create`?

snoopcheri (approved):

plugin/build.gradle.kts
pshevche: Hmm, reverted. Initially, the `compatibility` property could not be found if I used `register`, but now it works. It  may have been an issue with the sync

